PR #71, opened 2016-06-16
feat(select) implement optional select footer component

escarre:
This branch adds an optional footer to the novo-select element that adds a new item to the list. The footer will be displayed if a footerConfig object is passed in; that footerConfig should contain the label for the footer (the default display), the placeholder for the input (displayed when footer is open), and the onSave function. 

Primarily the Select element. Passed the footerConfig down through the FormInput --> SelectInput. Updated the SelectDemo to show an example.

##### **Reviewers**
* @jgodi 

### DIFF
--- a/src/elements/select/Select.js
+++ b/src/elements/select/Select.js
@@ -6,16 +6,26 @@ import { KeyCodes } from './../../utils/key-codes/KeyCodes';
 @Component({
     selector: 'novo-select',
     directives: [COMMON_DIRECTIVES],
-    inputs: ['options', 'placeholder', 'readonly'],
+    inputs: ['options', 'placeholder', 'readonly', 'footerConfig'],
     outputs: ['onSelect'],
     template: `
         <button (click)="toggleActive($event)" tabIndex="-1" type="button" [ngClass]="{empty: empty}">{{selected.label}}<i class="bhi-collapse"></i></button>
-        <ul class="novo-select-list" tabIndex="-1">
+        <ul class="novo-select-list" tabIndex="-1" [ngClass]="{footer: footerConfig}">
             <ng-content></ng-content>
             <li *ngFor="let option of options; let i = index" [ngClass]="{active: option.active}" (click)="onClickOption(option, i)" [attr.data-automation-value]="option.label">
               <span>{{option.label}}</span>
               <i *ngIf="option.active" class="bhi-check"></i>
             </li>
+            <div *ngIf="footerConfig" class="select-footer">
+                <button  *ngIf="!footer.open" (click)="toggleFooter($event); false" tabIndex="-1" type="button" class="footer"><i class="bhi-add-thin"></i>{{footerConfig.label}}</button>
+                <div *ngIf="footer.open" [ngClass]="{active: footer.open}">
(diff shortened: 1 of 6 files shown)


src/elements/select/_Select.scss
```diff
@@ -132,4 +138,103 @@ novo-select {
...
+    .select-footer {
+        button.footer  {
+            text-transform: uppercase;
+            color: $positive;
+            position: relative;
+            text-align: left;
+            cursor: pointer;
+            height: 48px;
+            margin: 0;
+            padding: 10px 16px 0px 10px;
```
krsween: Useless note: you don't need units on zero values in css.

src/elements/select/_Select.scss
```diff
@@ -132,4 +138,103 @@ novo-select {
...
+            }
+
+            input {
+                display: flex;
+                justify-content: space-between;
+                align-items: center;
+                background-color: transparent;
+                border: none;
+                border-bottom: 1px solid rgba(0, 0, 0, 0.12);
+                color: rgba(0, 0, 0, 0.73);
```
krsween: Pro-tip: you can also `rgba(black, .73)` with all CSS colors

src/elements/select/Select.js
```diff
@@ -6,16 +6,26 @@ import { KeyCodes } from './../../utils/key-codes/KeyCodes';
...
     template: `
         <button (click)="toggleActive($event)" tabIndex="-1" type="button" [ngClass]="{empty: empty}">{{selected.label}}<i class="bhi-collapse"></i></button>
-        <ul class="novo-select-list" tabIndex="-1">
+        <ul class="novo-select-list" tabIndex="-1" [ngClass]="{footer: footerConfig}">
             <ng-content></ng-content>
             <li *ngFor="let option of options; let i = index" [ngClass]="{active: option.active}" (click)="onClickOption(option, i)" [attr.data-automation-value]="option.label">
               <span>{{option.label}}</span>
               <i *ngIf="option.active" class="bhi-check"></i>
             </li>
+            <div *ngIf="footerConfig" class="select-footer">
```
jgodi: Footer should just be another <li> tag.

src/elements/select/Select.js
```diff
@@ -6,16 +6,26 @@ import { KeyCodes } from './../../utils/key-codes/KeyCodes';
...
+                <button  *ngIf="!footer.open" (click)="toggleFooter($event); false" tabIndex="-1" type="button" class="footer"><i class="bhi-add-thin"></i>{{footerConfig.label}}</button>
+                <div *ngIf="footer.open" [ngClass]="{active: footer.open}">
+                    <input autofocus type="text" [placeholder]="footerConfig.placeholder" [attr.id]="name" autocomplete="false" [(ngModel)]="footer.value" [ngClass]="{invalid: !footer.valid}"/>
+                    <footer>
+                        <button (click)="toggleFooter($event, false)">Cancel</button>
```
jgodi: The "Save/Cancel" should be added to the https://github.com/bullhorn/novo-elements/blob/master/src/services/novo-label-service.js, then you can inject that into Select and use `{{labels.save}}` instead of the plain text. This is used to localize it else where.

jgodi: Or you can get around that by having the save/cancel come from the config too.

src/elements/select/Select.js
jgodi: When toggling the "ADD NEW" the footer renders off screen, making me scroll down to save. Possible to not have it scroll when showing the input?

PR #129, opened 2016-08-02
feat(TipWell): adding new TipWell feature

krsween:
A TipWell can be used to display an in-line contextual help window for features that may confuse average users.

Added a new component, it's demo, and associated readme to novo-elements

##### **Reviewers**
* @jgodi 

### DIFF
--- a/src/elements/tip-well/TipWell.js
+++ b/src/elements/tip-well/TipWell.js
@@ -1,0 +1,66 @@
+// NG2
+import { Component } from '@angular/core';
+// App
+import { NOVO_BUTTON_ELEMENTS } from '../button';
+
+@Component({
+    selector: 'novo-tip-well',
+    inputs: [
+        'name',
+        'tip',
+        'buttonText'
+    ],
+    directives: [
+        NOVO_BUTTON_ELEMENTS
+    ],
+    template: `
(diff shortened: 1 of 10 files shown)


src/elements/tip-well/TipWell.js
```diff
@@ -0,0 +1,66 @@
...
+export class TipWell {
+    constructor() {
+        this.isActive = true;
+        // Check if localStorage is enabled
+        this.isLocalStorageEnabled = (() => {
+            let isEnabled = false;
+            if (typeof localStorage === 'object') {
+                try {
+                    localStorage.setItem('lsTest', 1);
+                    localStorage.removeItem('lsTest');
```
amrutha-rajiv: just a thought @krsween, all tip wells may not need buttons and all tip wells' buttons may not be doing the same things so it will be great if in the future we can make the button optional and add a call back option for what the button would do so that the button action can be controlled by the user of the tip well.

krsween:
That's a fair point. I couldn't think of a place where that existed in our project, but for the sake of flexibility we should roadmap some of these enhancements.

amrutha-rajiv:
I'll add it to our issue list

PR #135, opened 2016-08-03
feat(Table): changed Date filtering to allow multiple or single selection

MagicAardvark:
allows date filters to be toggled between multiple and single selection. with multiple=true property on meta data.

Table.js and spec file.

Changed includes to indexOf check due to PhantomJS not supporting includes, no shim in polyfills

##### **Reviewers**
* @bvkimball @jgodi @krsween 

### DIFF
--- a/src/elements/table/Table.js
+++ b/src/elements/table/Table.js
@@ -260,7 +260,7 @@ export class NovoTable {
      */
     onFilterClick(column, filter) {
         if (Array.isArray(column.filter) && column.multiple) {
-            if (column.filter.indexOf(filter) > -1) {
+            if (~column.filter.indexOf(filter)) {
                 // Remove filter
                 column.filter.splice(column.filter.indexOf(filter), 1);
 
--- a/src/elements/table/Table.js
+++ b/src/elements/table/Table.js
@@ -259,8 +259,8 @@ export class NovoTable {
      * @param filter
      */
     onFilterClick(column, filter) {
-        if (Array.isArray(column.filter)) {
-            if (column.filter.includes(filter)) {
+        if (Array.isArray(column.filter) && column.multiple) {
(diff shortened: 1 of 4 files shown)


src/elements/table/Table.js
```diff
@@ -259,8 +259,8 @@ export class NovoTable {
      * @param filter
      */
     onFilterClick(column, filter) {
-        if (Array.isArray(column.filter)) {
-            if (column.filter.includes(filter)) {
+        if (Array.isArray(column.filter) && column.multiple) {
```
jgodi: Does this break the other table filters that use select boxes and were multiple by default?

src/elements/table/Table.js
```diff
@@ -259,8 +259,8 @@ export class NovoTable {
...
+            if (column.filter.indexOf(filter) > -1) {
```
krsween: Can you tilde this instead? I think it's a preference:

`if (column.filter.indexOf(filter) > -1) {`

`if (~column.filter.indexOf(filter)) {`

[Fiddle](https://jsfiddle.net/krsween/kf7q5cjL/)

src/elements/table/Table.js
krsween: Not to pile on, but whatever the decision for multiple by default or single, can we call that out in the readme? 

https://github.com/bullhorn/novo-elements/blob/master/src/elements/table/README.md

jgodi: Just tested it and it does change the current functionality for selection based filters. Check the Table -> Basic -> Filter Status Column. This was multiple by default.

I am fine with making it single by default and making the `multiple=true`, but you would have to go through all the columns that are currently inside all Novo apps and make sure they get `multiple=true` if they were before.

@krsween @bvkimball thoughts?

PR #152, opened 2016-08-16
F/delp

krsween:
When you select an element in picker, then focus that instance of the picker, then hit backspace, the label of the picker changes, but the actual value does not.

Picker will now clear when there is a value selected and you backspace the term of the picker.  Linguistically, this sounds a lot more complicated than it is.

##### **Reviewers**
* @jgodi 

### DIFF
--- a/src/elements/table/Table.js
+++ b/src/elements/table/Table.js
@@ -121,7 +121,7 @@ export class NovoTableHeader {
                                             <button theme="dialogue" color="negative" icon="times" (click)="onFilterClear(column)" *ngIf="column.filter">{{ labels.clear }}</button>
                                         </div>
                                     </item>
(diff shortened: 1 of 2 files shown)


src/elements/table/Table.js
```diff
@@ -121,7 +121,7 @@ export class NovoTableHeader {
                                             <button theme="dialogue" color="negative" icon="times" (click)="onFilterClear(column)" *ngIf="column.filter">{{ labels.clear }}</button>
                                         </div>
                                     </item>
-                                    <item [ngClass]="{ active: isFilterActive(column, option) }" *ngFor="let option of column.options" (click)="onFilterClick(column, option)" [showAfterSelect]="option.range" [attr.data-automation-id]="option">
+                                    <item [ngClass]="{ active: isFilterActive(column, option) }" *ngFor="let option of column.options" (click)="onFilterClick(column, option)" [showAfterSelect]="option.range" [attr.data-automation-id]="(option?.label || option)">
```
krsween: This should already be merged...

jgodi: Odd

PR #164, opened 2016-08-25
fix(forms): Fixing initial values for the updated values and validation

jgodi:
Fixing some bugs with forms

### DIFF
--- a/src/elements/form/Control.js
+++ b/src/elements/form/Control.js
@@ -44,7 +44,7 @@ export class NovoControlElement extends OutsideClick {
     }
 
     get isDirty() {
-        return this.form.controls[this.control.key].dirty;
+        return this.form.controls[this.control.key].dirty || this.control.dirty;
     }
 
     formatDateTimeValue(event, part) {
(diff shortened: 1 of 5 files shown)


src/elements/form/Control.js
```diff
@@ -44,7 +44,7 @@ export class NovoControlElement extends OutsideClick {
     }
 
     get isDirty() {
-        return this.form.controls[this.control.key].dirty;
+        return this.form.controls[this.control.key].dirty || this.control.dirty;
```
jgodi: Never showed initial validation failures

src/elements/form/DynamicForm.js
```diff
@@ -81,7 +81,7 @@ export class NovoDynamicFormElement {
         return this.form ? this.form.valid : false;
     }
 
-    get updatedValues() {
+    updatedValues() {
```
jgodi: Was called literally every dirty check.. now this isn't referenced in the templates in the app and just invoked

src/elements/form/controls/BaseControl.js
```diff
@@ -13,6 +13,7 @@ export class BaseControl {
         this.controlType = config.controlType || '';
         this.placeholder = config.placeholder || '';
         this.config = config.config || null;
+        this.dirty = !!config.value;
```
jgodi: Mark it as dirty if we have a value so `updateValues` stays right

PR #226, opened 2016-10-26
refactor(forms): change FormUtils to include Placement

krsween:
Placements don't get the entity pickers.

Added Placements to the TO_ONE and TO_MANY types that get pickers and chips.

##### **Reviewers**
* @jgodi 

### DIFF
--- a/src/elements/form/FormUtils.js
+++ b/src/elements/form/FormUtils.js
@@ -90,9 +90,7 @@ export class FormUtils {
         }
 
         // Overrides
-        if (type === 'checkbox' && field.options) {
-            type = 'checklist';
-        } else if (type === 'picker' && field.multiValue) {
+        if (type === 'picker' && field.multiValue) {
             type = 'chips';
         } else if (type === 'entitypicker' && field.multiValue) {
             type = 'entitychips';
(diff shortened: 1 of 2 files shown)


krsween:
@jgodi hold off on this PR. I want to at least push a TDD stub.

krsween:
Provided these specs are all good. This is stable.

src/elements/form/FormUtils.js
```diff
@@ -91,9 +90,7 @@ export class FormUtils {
         }
 
         // Overrides
-        if (type === 'checkbox' && field.options) {
```
krsween: @jgodi There is no setter in this code block that sets `type` equal to 'checkbox', so this check would never be true.

krsween:
Pretty sure **coveralls** is lying as I just added 180 lines of tests where there were previously 0.

jgodi:
But before this file was not reported coverage on since there was not a spec :P

src/elements/form/FormUtils.js
jgodi: This was from the old-old-old FormUtils before the first major Form Refactor

Issue #232, opened 2016-10-31
[Button] - Disabled state displays as active, not disabled


escarre:
_**Note: for support questions, please use one of these channels:**_ [Chat: BullhornOSS Gitter](https://gitter.im/bullhorn/Open-Source)

##### **What is the current behavior?**
Disabled buttons display as active.

##### **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via https://plnkr.co or similar. (TODO - make a starter plnkr)

See current demo. In the Primary buttons section, the Submit button has the disabled attribute, but it is in an active state.
![screen shot 2016-10-31 at 10 14 23 am](https://cloud.githubusercontent.com/assets/7002678/19859335/e9fc95ea-9f52-11e6-8bd1-8b0aa964f7c1.png)

##### **What is the expected behavior?**
Disabled buttons should be disabled. :)

##### **Please tell us about your environment:**

- Angular version: 2.0.0-RC.X
- Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

##### **Other information** 
_(e.g. detailed explanation, screenshots, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)_
This was introduced when the loading attribute was added.

PR #233, opened 2016-10-31
Bug (Button): Fixes buttons so that they apply disabled attribute whe…

escarre:
This PR unties the disabled attribute from a component's loading input so that buttons will apply as disabled when intended. (See issue #232)

It applies the disabled css within the loading attribute.  

- Untied disabled attribute from loading input
- Added disabled css to loading

##### **Reviewers**
* @jgodi 

…n disabled - issue #232

Applies disabled styles as part of the loading attribute, rather than using the host to set the disabled attribute when loading.

### DIFF
--- a/src/elements/button/Button.js
+++ b/src/elements/button/Button.js
@@ -8,7 +8,6 @@ import { Component } from '@angular/core';
         '[attr.theme]': 'theme',
         '[attr.color]': 'color',
         '[attr.icon]': 'icon',
-        '[disabled]': 'loading',
         '[attr.loading]': 'loading'
     },
     template: `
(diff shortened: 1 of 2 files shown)


src/elements/button/Button.js
```diff
@@ -8,7 +8,6 @@ import { Component } from '@angular/core';
         '[attr.theme]': 'theme',
         '[attr.color]': 'color',
         '[attr.icon]': 'icon',
-        '[disabled]': 'loading',
```
jgodi: We want the button to be disabled when 'loading' is set to true still.

PR #246, opened 2016-11-14
fix:(address fields): Show and populate address fields

steveerfalot:
Show and populate address fields

I changed the Address `writeValue` function to actually populate the address field, and changed formUtils to show the address field

##### **Reviewers**
* @jgodi 
* @bvkimball 

### DIFF
--- a/src/elements/form/extras/address/Address.js
+++ b/src/elements/form/extras/address/Address.js
@@ -1,8 +1,8 @@
 // NG2
-import { Component, forwardRef } from '@angular/core';
+import { Component, forwardRef, Input } from '@angular/core';
 import { NG_VALUE_ACCESSOR, ControlValueAccessor } from '@angular/forms';
 // APP
-import { getCountries, getStates, findByCountryName } from '../../../../utils/countries/Countries';
+import { getCountries, getStates, findByCountryName, findByCountryId } from '../../../../utils/countries/Countries';
 
 // Value accessor for the component (supports ngModel)
 const ADDRESS_VALUE_ACCESSOR = {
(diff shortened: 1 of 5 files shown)


src/elements/form/extras/address/Address.js
```diff
@@ -73,17 +82,25 @@ export class NovoAddressElement implements ControlValueAccessor {
...
+        setTimeout(() => {
+            if (!model) {
+                this.model = {
+                    countryID: 1,
+                    countryName: 'United States'
+                };
+            } else {
+                model.countryName = findByCountryId(model.countryID).name;
+                this.model = model;
+                let modelState = this.states.find(state => {
```
jgodi: This assumes all states are in the full name `Missouri` vs. `MO`. There are instances where the data can be both. If `MO` was passed then the state would not match since `this.states` is `['Missouri']`

src/elements/form/extras/address/Address.js
```diff
@@ -73,17 +82,25 @@ export class NovoAddressElement implements ControlValueAccessor {
...
     writeValue(model:any):void {
-        if (!this.model || !this.model.length) {
-            this.model = {
-                countryID: 1,
-                countryName: 'United States'
-            };
-        } else {
-            this.model = model;
-        }
+        setTimeout(() => {
```
jgodi: Should not call `setTimeout` in the `writeValue` block.

src/elements/form/extras/address/Address.js
```diff
@@ -47,14 +51,19 @@ export class NovoAddressElement implements ControlValueAccessor {
     }
 
     onCountryChange(evt) {
-        this.model.countryName = evt;
-        let country = findByCountryName(this.model.countryName);
+        let country;
+        if (typeof evt === 'number') {
```
jgodi: Should go into `Helpers`, then `Helpers.isNumber`.

src/elements/form/extras/address/Address.js
```diff
@@ -28,6 +29,9 @@ export class NovoAddressElement implements ControlValueAccessor {
     countries:Array = getCountries();
 
     model;
+    constructor(changeDetectorRef:ChangeDetectorRef) {
```
jgodi: Constructor after property declarations please

src/elements/form/extras/address/Address.js
jgodi: The data that comes back for address for most entities is:

```
"address": {
"address1": "",
"address2": "",
"city": "Pleasanton",
"state": "CA",
"zip": "94588",
"countryID": 1,
"countryName": "United States                                                   ",
"countryCode": "US"
}
```

While, JobOrder/Opportunity are the only ones that look like

```
"address": {
"address1": "HP-8",
"city": "High Point",
"state": "North Carolina",
"zip": "27261",
"countryID": 1
}
```

In that case, the data is already there and this is doing unnecessary checks. We will still have to handle the state issue (see comment below).

There is a SAT Chat topic on fixing Address fields, since we have three types: Address, Address1, and AddressWithoutCountry.  

I think the solution should be a mixture of what is here and fixing REST.

Plus there is also a ticket in Fit&Finish (or the spreadsheet) to fix required fields and Address. Since sometimes only city/state is required and others everything is.

steveerfalot: Without the `setTimeout` we get property has changed after it was checked errors. Should we add a check to only set `countryName` if it doesn't exist?

src/elements/form/extras/address/Address.js
steveerfalot: The current state of `getStates` only returns the `name` of the state, not the `code`. 

```
export function getStates(name) {
    let country = COUNTRIES.find(c => c.name === name.trim());
    return country ? country.states.map(s => s.name) : [];
}
```

jgodi: Yes, sometimes the data comes back as `state: 'MO'`, which would not work with this code.

steveerfalot: So, should we also fix `getStates`?

src/elements/form/extras/address/Address.js
jgodi: Yes to the check. That error only occurs because you do the state stuff I think. If you were to use `this.model = Object.assign({ countryName, state }, model)` and not call into `this.onStateChange` it should not occur.

src/elements/form/extras/address/Address.js
bvkimball: To handle the states by Code i think we would want to add an option to the control that sets the state picker to use the `code` or the 'name` field for the states.

steveerfalot: Should that be a later fix, or does it need to be addressed now?

PR #249, opened 2016-11-15
Fix logic to respect when column sorting is false

adrianossi:
Fix table and sorting logic so that columns marked sorting: false do not display sorting features.

Table.js
_Table.scss
ThSortable.js

##### **Reviewers**
* @user

### DIFF
--- a/src/elements/table/extras/th-sortable/ThSortable.js
+++ b/src/elements/table/extras/th-sortable/ThSortable.js
@@ -21,7 +21,7 @@ export class ThSortable {
             event.preventDefault();
         }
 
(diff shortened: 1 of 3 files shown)


src/elements/table/extras/th-sortable/ThSortable.js
```diff
@@ -21,7 +21,7 @@ export class ThSortable {
             event.preventDefault();
         }
 
-        if (this.config && this.column && this.config.sorting !== false && this.column.sort !== false) {
+        if (this.config && this.column && this.config.sorting !== false && this.column.sorting !== false) {
```
jgodi: Still want sort here? The sort is 'ASC' or 'DESC', there is already 'sorting' check

PR #285, opened 2016-12-12
fix(chips): Picker results for chips will now stay open when selecting.

msparkman:


jgodi:
Can you flip this? This will break all the current pickers and make then stay open. Shouldn't it be the default to close and then you say `keepOpenOnSelect` or something to make it stay open?

jgodi:
Because this feature will break automation if it is not something to opt into. Then inside the app, opt in to everywhere UX wants and then fix the automation :)

msparkman:
Yes, I can go flip it. That does makes sense.

PR #293, opened 2016-12-14
fix(table): add back in custom sorting/filtering functions

escarre:
Canvas-UI uses a custom sorting/filtering function to load sorted/filtered reports. This PR adds back in that functionality.

Table.ts

##### **Reviewers**
* @jgodi 

### DIFF
--- a/src/elements/table/Table.ts
+++ b/src/elements/table/Table.ts
@@ -362,40 +362,39 @@ export class NovoTableElement implements DoCheck {
             // Array of filters
             const filters = this.columns.filter(col => col.filter && col.filter.length);
             if (filters.length) {
-                if (Helpers.isFunction(this.config.filtering)) {
-                    this.config.filtering();
-                    return;
-                } else {
-                    let query = {};
-                    for (const column of filters) {
-                        if (Array.isArray(column.filter)) {
-                            // The filters are an array (multi-select), check value
-                            if (column.type && column.type === 'date' && column.filter.filter(fil => fil.range).length > 0) {
-                                query[column.name] = column.filter.map(f => {
-                                    return {
-                                        min: f.value ? new Date(f.value.startDate).getTime() : 0,
-                                        max: f.value ? new Date(f.value.endDate).getTime() : 0
-                                    };
-                                })[0];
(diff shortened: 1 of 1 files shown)


src/elements/table/Table.ts
```diff
@@ -362,37 +362,42 @@ export class NovoTableElement implements DoCheck {
...
-                                    min: f.min ? Date.now() + (f.min * (24 * 60 * 60 * 1000)) : Date.now(),
-                                    max: f.max ? Date.now() + (f.max * (24 * 60 * 60 * 1000)) : Date.now()
-                                };
-                            })[0];
-                        } else {
-                            let options = column.filter;
-                            // We have an array of {value: '', labels: ''}
-                            if (options[0].value || options[0].label) {
-                                options = column.filter.map(opt => opt.value);
+                if (Helpers.isFunction(this.config.filtering)) {
```
bvkimball: Move this if to after the `query` is processed and send the query as a parameter.

src/elements/table/Table.ts
```diff
@@ -362,37 +362,42 @@ export class NovoTableElement implements DoCheck {
...
+                            query[column.name] = column.filter;
                         }
-                    } else {
-                        query[column.name] = column.filter;
                     }
+                    this._dataProvider.filter = query;
                 }
-                this._dataProvider.filter = query;
             } else {
                 this._dataProvider.filter = {};
```
bvkimball: this should be 
```
if (Helpers.isFunction(this.config.filtering)) {
      @bvkimball
bvkimball just now Member
Move this if to after the query is processed and send the query as a parameter.
 @bvkimball	Reply…
if (Helpers.isFunction(this.config.filtering)) {
    this.config.filtering(query);
} else {
    this._dataProvider.filter = query;
}

File: src/elements/table/Table.ts [hunk 3]
```diff
@@ -443,6 +447,9 @@ export class NovoTableElement implements DoCheck {
         this.currentSortColumn = column;
 
         if (column) {
+            if (Helpers.isFunction(this.config.sorting)) {
+                this.config.sorting();
+            }
```
bvkimball: else?

PR #335, opened 2017-01-09
BH-46505 - making state no longer a required field

ncal27:
_REPLACE ME WITH A SHORT DESCRIPTION_

Removed validation check for state field

##### **Reviewers**
* @user


jgodi:
Why would we not just update the field mappings for the address fields and then make the address control look at the field mappings for required or not, then the validator can be updated to the same.

PR #344, opened 2017-01-11
feat(dropdown): adding typeahead for dropdowns

mcarver87:
added a typeahead for country dropdowns

##### **Reviewers**
* @user


src/elements/select/Select.ts
```diff
@@ -62,9 +64,15 @@ export class NovoSelectElement extends OutsideClick implements OnInit, OnChanges
     };
     onModelTouched: Function = () => {
     };
+    filterTerm: string;
 
     constructor(element: ElementRef, public labels: NovoLabelService) {
         super(element);
+        this.filterTerm = '';
+
+        let keydown = Observable.fromEvent(document, 'keydown')
```
jgodi: This puts a global key handler on each select we have on the page, 10 selects - 10 handlers and they never get removed

jgodi: Is this just to clear the filer after 500ms?

PR #351, opened 2017-01-13
chore(Form): added label property to form controls to be accessible v…

krsween:
Needed `label` value from form control.

Added label property to form controls to be accessible via controller (for i18n).

##### **Reviewers**
* @jgodi 

### DIFF
--- a/src/elements/form/FormUtils.ts
+++ b/src/elements/form/FormUtils.ts
@@ -27,10 +27,12 @@ export class NovoFormControl extends FormControl {
     hidden: boolean;
     required: boolean;
     initialValue: any;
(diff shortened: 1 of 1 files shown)


jgodi:
if you want the changing of this label to be reflected in the form itself, you will need to update:

https://github.com/bullhorn/novo-elements/blob/master/src/elements/form/Control.ts#L17

and 

https://github.com/bullhorn/novo-elements/blob/master/src/elements/form/Control.ts#L29

to be like `{{ form.controls[control.key].label }}`

since it will only show the label off the BaseControl and not the NovoFormControl

PR #360, opened 2017-01-19
Implement Character Count and error messages for fields with maxLength 

ncal27:
_REPLACE ME WITH A SHORT DESCRIPTION_

Added maxLength check and character count for fields with it

##### **Reviewers**
* @user

### DIFF
--- a/src/elements/form/Control.ts
+++ b/src/elements/form/Control.ts
@@ -35,7 +35,7 @@ import { Helpers } from './../../utils/Helpers';
                     <!--Text-based Inputs-->
                     <!--TODO prefix/suffix on the control-->
                     <div class="novo-control-input-container novo-control-input-with-label" *ngSwitchCase="'textbox'">
-                        <input *ngIf="control.type !== 'number'" [formControlName]="control.key" [id]="control.key" [type]="control.type" [placeholder]="control.placeholder" (input)="emitChange($event)" [maxlength]="control.maxlength" (focus)="handleFocus($event)" (blur)="handleBlur($event)">
+                        <input *ngIf="control.type !== 'number'" [formControlName]="control.key" [id]="control.key" [type]="control.type" [placeholder]="control.placeholder" (input)="emitChange($event)" (keypress)="checkMaxLength($event)" [maxlength]="control.maxlength" (focus)="handleFocus($event)" (blur)="handleBlur($event)">
                         <input *ngIf="control.type === 'number'" [formControlName]="control.key" [id]="control.key" [type]="control.type" [placeholder]="control.placeholder" (keypress)="restrictKeys($event)" (input)="emitChange($event)" [maxlength]="control.maxlength" (focus)="handleFocus($event)" (blur)="handleBlur($event)" step="any">
                         <label class="input-label" *ngIf="control.subType === 'currency'">{{control.currencyFormat}}</label>
                         <label class="input-label" *ngIf="control.subType === 'percentage'">%</label>
--- a/src/elements/form/Control.ts
+++ b/src/elements/form/Control.ts
@@ -99,7 +99,8 @@ import { Helpers } from './../../utils/Helpers';
             <div class="error-message">
                 <span *ngIf="isDirty && errors?.required">{{control.label | uppercase}} is required</span>
                 <span *ngIf="isDirty && errors?.minlength">{{control.label | uppercase}} is required to be a minimum of {{ control.minlength }} characters</span>
-                <span *ngIf="isDirty && errors?.maxlength">{{control.label | uppercase}} is required to be a maximum of {{ control.maxlength }} characters</span>
+                <span *ngIf="isDirty && maxLengthMet && !maxLength">Sorry, you have reached the maximum character count of {{ control.maxlength }} for this field</span>
+                <span *ngIf="errors?.maxlength">Sorry, you have exceeded the maximum character count of {{ control.maxlength }} for this field</span>
                 <span *ngIf="isDirty && errors?.invalidEmail">{{control.label | uppercase}} requires a valid email (ex. abc@123.com)</span>
                 <span *ngIf="isDirty && errors?.invalidAddress">{{control.label | uppercase}} requires all fields filled out</span>
                 <span *ngIf="isDirty && (errors?.integerTooLarge || errors?.doubleTooLarge)">{{control.label | uppercase}} is too large</span>
@@ -134,6 +135,7 @@ export class NovoControlElement extends OutsideClick implements OnInit, OnDestro
     private _focusEmitter: EventEmitter<FocusEvent> = new EventEmitter<FocusEvent>();
     private _focused: boolean = false;
     formattedValue: string = '';
+    maxLengthMet: boolean = false;
 
     constructor(element: ElementRef, public labels: NovoLabelService, private toast: NovoToastService) {
         super(element);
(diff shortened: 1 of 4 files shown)


src/elements/form/Control.ts
```diff
@@ -95,11 +95,15 @@ import { Helpers } from './../../utils/Helpers';
...
             </div>
+            <div class="text-field-character-limit" [attr.data-automation-id]="text-area-character-count" *ngIf="_focused && control.maxlength && (control.controlType=='text-area' || control.controlType=='textbox')">
+                Character Count: {{ characterCount }}/{{ control.maxlength }}
+            </div>
             <!--Error Message-->
             <div class="error-message">
                 <span *ngIf="isDirty && errors?.required">{{control.label | uppercase}} is required</span>
                 <span *ngIf="isDirty && errors?.minlength">{{control.label | uppercase}} is required to be a minimum of {{ control.minlength }} characters</span>
-                <span *ngIf="isDirty && errors?.maxlength">{{control.label | uppercase}} is required to be a maximum of {{ control.maxlength }} characters</span>
+                <span *ngIf="isDirty && maxLengthMet && _focused && !errors?.maxlength">Sorry, you have reached the maximum character count of {{ control.maxlength }} for this field</span>
```
jgodi: `_focused` is private, use `focused`

src/elements/form/Control.ts
```diff
@@ -95,11 +95,15 @@ import { Helpers } from './../../utils/Helpers';
                     <novo-quick-note *ngSwitchCase="'quick-note'" [formControlName]="control.key" [placeholder]="control.placeholder" [config]="control.config" (change)="modelChange($event)"></novo-quick-note>
                 </div>
             </div>
+            <div class="text-field-character-limit" [attr.data-automation-id]="text-area-character-count" *ngIf="_focused && control.maxlength && (control.controlType=='text-area' || control.controlType=='textbox')">
```
jgodi: Styling for this is off, UX wanted it to read `LENGTH/MAX_LENGTH` and be over on the right side, on the same line as the error.

screenshot 2017-01-19 14 38 38

should resemble

screenshot 2017-01-19 14 41 16

and turn red when in error too like

screenshot 2017-01-19 14 41 24

with the error showing up to the left and the count to the right, all on the same line

jgodi: Also `focused` here instead of the private variable

src/elements/form/Control.ts
```diff
@@ -35,13 +35,13 @@ import { Helpers } from './../../utils/Helpers';
                     <!--Text-based Inputs-->
                     <!--TODO prefix/suffix on the control-->
                     <div class="novo-control-input-container novo-control-input-with-label" *ngSwitchCase="'textbox'">
-                        <input *ngIf="control.type !== 'number'" [formControlName]="control.key" [id]="control.key" [type]="control.type" [placeholder]="control.placeholder" (input)="emitChange($event)" [maxlength]="control.maxlength" (focus)="handleFocus($event)" (blur)="handleBlur($event)">
+                        <input *ngIf="control.type !== 'number'" [formControlName]="control.key" [id]="control.key" [type]="control.type" [placeholder]="control.placeholder" (input)="emitChange($event)" (keypress)="checkMaxLength($event)" [maxlength]="control.maxlength" (focus)="handleFocus($event)" (blur)="handleBlur($event)">
```
jgodi: Don't need keypress, you have the call inside emitChange, so you are calling it twice. Can remove the keypress event listenter.

src/elements/form/Control.ts
```diff
@@ -35,13 +35,13 @@ import { Helpers } from './../../utils/Helpers';
...
                         <label class="input-label" *ngIf="control.subType === 'currency'">{{control.currencyFormat}}</label>
                         <label class="input-label" *ngIf="control.subType === 'percentage'">%</label>
                     </div>
                     <!--TextArea-->
-                    <textarea *ngSwitchCase="'text-area'" [name]="control.key" [attr.id]="control.key" [placeholder]="control.placeholder" [formControlName]="control.key" (input)="resizeTextArea($event)" (input)="emitChange($event)" [maxlength]="control.maxlength"></textarea>
+                    <textarea *ngSwitchCase="'text-area'" [name]="control.key" [attr.id]="control.key" [placeholder]="control.placeholder" [formControlName]="control.key" (input)="resizeTextArea($event)" (keypress)="checkMaxLength($event)" (input)="emitChange($event)" [maxlength]="control.maxlength"></textarea>
```
jgodi: Don't need keypress, you have the call inside emitChange, so you are calling it twice. Can remove the keypress event listenter.

src/elements/form/Control.ts
jgodi: I would put this in the same error message div and use flex box to position them, using different colors for the count if its not in error and that.

src/elements/form/Control.ts
```diff
@@ -251,37 +259,51 @@ export class NovoControlElement extends OutsideClick implements OnInit, OnDestro
...
         // Reset the height
         let height = event.target.value.length > 0 ? `${event.target.scrollHeight}px` : '2rem';
         event.target.style.height = '';
         event.target.style.height = height;
     }
 
+    checkMaxLength(event) {
+        if (this.control && this.control.maxlength) {
+            this.characterCount = event.target.value.length;
+            if (event.target.value.length >= this.control.maxlength) {
```
jgodi: `this.maxLengthMet = event.target.value.length >= this.control.maxlength`

src/elements/form/Control.ts
```diff
@@ -251,37 +259,51 @@ export class NovoControlElement extends OutsideClick implements OnInit, OnDestro
...
+            }
+        }
+    }
+
     modelChange(value) {
         if (!value) {
             this._focused = false;
         }
         this.change.emit(value);
+        this.checkMaxLength(value);
```
jgodi: `modelChange` is only used on Picker/Tiles/Chips, don't think you need to check it here. You only care about textbox and textarea

src/elements/form/Control.ts
ncal27: I talked to mary a bit more about this and left some comments on the ticket based on that 9 I will link it for reference), but I will look into why the character count is in the middle, cause its not supposed to be there. https://jira.bullhorn.com/browse/BH-46921

jgodi: If that is so, it will need to be localized with the `labels` that are injected in via `NovoLabelService`

jgodi: Probably all of those errors need to be as well if you have time

ncal27: It doesnt make sense to me to put it in the error message div, cause its not an error message, its the character count

src/elements/form/Control.ts
```diff
@@ -99,7 +99,11 @@ import { Helpers } from './../../utils/Helpers';
             <div class="error-message">
                 <span *ngIf="isDirty && errors?.required">{{control.label | uppercase}} is required</span>
                 <span *ngIf="isDirty && errors?.minlength">{{control.label | uppercase}} is required to be a minimum of {{ control.minlength }} characters</span>
-                <span *ngIf="isDirty && errors?.maxlength">{{control.label | uppercase}} is required to be a maximum of {{ control.maxlength }} characters</span>
+                <span class="character-limit-error" *ngIf="isDirty && maxLengthMet && focused && !errors?.maxlength">Sorry, you have reached the maximum character count of {{ control.maxlength }} for this field</span>
+                <span class="character-limit-error" *ngIf="errors?.maxlength">Sorry, you have exceeded the maximum character count of {{ control.maxlength }} for this field</span>
+                <span class="text-field-character-limit" [attr.data-automation-id]="text-area-character-count" *ngIf="focused && control.maxlength && (control.controlType=='text-area' || control.controlType=='textbox')">
```
jgodi: `[attr.data-automation-id]="text-area-character-count"` there is no property `text-area-character-count` on the class, did you want it to be just a string? if so then you will have to escape it`[attr.data-automation-id]="'text-area-character-count'"`

jgodi: This should also always appear to the right even without an error.

screenshot 2017-01-20 10 09 51
screenshot 2017-01-20 10 09 58
screenshot 2017-01-20 10 10 05

it should always be on the bottom right even with error and without error

PR #366, opened 2017-01-20
fix(table) fix issue with multiple columns having sort enabled

jeremy-dobler:
Issue occurring in Canvas where clicking filter and/or sorts really fast would cause the state of the table to be bad. Multiple columns would be registered as sorted.

Now onSortChange will remove the sort from every other column except for the column that is currently getting sorted.

##### **Reviewers**
* @user

* [x ] Changes are limited to a single goal (no scope creep)
* [x ] Code can be automatically merged (no conflicts)
* [x ] Code follows the standards laid out in the [contributing guide](https://github.com/bullhorn/novo-elements/blob/master/CONTRIBUTING.md)
* [x ] Passes all existing automated tests
* [x ] New functions are documented (with a description, list of inputs, and expected output)
* [x ] Visually tested in supported browsers and devices

### DIFF
--- a/src/elements/table/Table.ts
+++ b/src/elements/table/Table.ts
@@ -478,10 +478,13 @@ export class NovoTableElement implements DoCheck {
      * @param newSortColumn
      */
     onSortChange(column) {
-        if (this.currentSortColumn && this.currentSortColumn !== column) {
-            this.currentSortColumn.sort = null;
-        }
         this.currentSortColumn = column;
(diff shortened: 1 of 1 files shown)


jeremy-dobler:
reworded commit from "feat" to "fix" to match guidelines for the type of commit

jgodi:
I squash and merge anyways, so the I alter the commit message then if it doesn't match.

I will let @bvkimball review this since he was doing a lot around here.

PR #379, opened 2017-01-31
feat(Table): Table filtering should be disabled for all columns if set in config

MagicAardvark:
##### **Description**
Table filtering should be disabled for all columns if set in config
config = { filtering: false }

##### **Reviewers**
* @jgodi
* @more


jgodi:
There was a reason why we didnt do this, let me try to remember.

PR #392, opened 2017-02-07
BH-47871 added another parameter to callback function in category drop…

DrewDudley:
…down footer

##### **Reviewers**
* @jgodi
* @more

### DIFF
--- a/src/elements/category-dropdown/CategoryDropdown.ts
+++ b/src/elements/category-dropdown/CategoryDropdown.ts
@@ -36,7 +36,7 @@ import { Helpers } from './../../utils/Helpers';
                 </novo-nav-content>
             </novo-nav-outlet>
             <footer *ngIf="footer" class="novo-category-dropdown-footer-align-{{ footer.align || 'right' }}">
-                <a *ngFor="let link of footer.links" (click)="link.callback($event)">{{ link.label }}</a>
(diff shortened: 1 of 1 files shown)


src/elements/category-dropdown/CategoryDropdown.ts
```diff
@@ -36,7 +36,7 @@ import { Helpers } from './../../utils/Helpers';
                 </novo-nav-content>
             </novo-nav-outlet>
             <footer *ngIf="footer" class="novo-category-dropdown-footer-align-{{ footer.align || 'right' }}">
-                <a *ngFor="let link of footer.links" (click)="link.callback($event)">{{ link.label }}</a>
+                <a *ngFor="let link of footer.links" (click)="link.callback($event, this)">{{ link.label }}</a>
             </footer>
```
amrutha-rajiv: I would recommend doing something like this instead 

```
 executeClickCallback(event, link) {
        link.callback(event);
        // Close, if input is set
        if (this.closeOnSelect) {
            this.toggleActive();
        }
    }
    (click)="executeClickCallback(link, $event)
```

DrewDudley: No problem. Just curious though, is there a scenario in which a user would not want the dropdown to close when they click on a footer element?

amrutha-rajiv: at that point that can set `closeOnSelect=false` in the picker config

PR #396, opened 2017-02-08
BH-45576 Added truncation to card titles

nathan-wilson-bh:
##### **Description**

Added truncation to card titles

##### **Reviewers**
* @jgodi
* @more

### DIFF
--- a/src/elements/card/Card.scss
+++ b/src/elements/card/Card.scss
@@ -25,19 +25,26 @@ novo-card {
             .title {
                 display: flex;
                 align-items: center;
+                height: 1.75em;
+                overflow: hidden;
 
                 i.bhi-move {
                     color: $light;
                     margin-right: 10px;
+                    height: 1.3em
                 }
 
                 h3 {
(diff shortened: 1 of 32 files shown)


src/elements/card/Card.scss
```diff
@@ -24,20 +24,28 @@ novo-card {
...
                     color: $light;
                     margin-right: 10px;
+                    height: 1.3em
                     cursor: pointer;
+
                 }
 
                 h3 {
                     padding: 0;
+                    display: -webkit-box;
```
jgodi: This only works in chrome then? `white-space: nowrap;text-overflow: ellipsis;`?

nathan-wilson-bh: Just tried that. The card container doesn't respect it's own size with that implementation. The card grows off of the screen instead of wrapping.

jgodi: This should be cross-browser, we shouldn't be putting in vendor specific CSS here.

Issue #333, opened 2017-01-09
[Forms] - Tooltips on form controls should be passed down to form element


asibilia:
##### **What is the current behavior?**
Currently there is no way to add a `tooltip` to form elements via the `form-control` templates. 

##### **What is the expected behavior?**
When you add a `tooltip` to a form control it should be passed down to the specific form element rather than remain on the top level control element.

##### **What is the motivation / use case for changing the behavior?**
Since the tooltips stay on the top level `form-control` instead of being passed down to the form element (input, tile, checkbox etc.) the positioning of the `tooltip` is relative to the entire container and gets positioned incorrectly. This often results in tooltips showing up outside of modals / slide outs because the `form-control` is usually full width.

##### **Current**
![tooltip error](https://cloud.githubusercontent.com/assets/5579814/21781238/388cefea-d674-11e6-8d84-40c6fc3fa119.png)

##### **Proposed**
![tooltip-correct](https://cloud.githubusercontent.com/assets/5579814/21781246/419edf6c-d674-11e6-9fa4-a67f4eaca324.png)

PR #401, opened 2017-02-12
F/form updates

amrutha-rajiv:
##### **Description**

screen shot 2017-02-12 at 5 35 45 pm
screen shot 2017-02-12 at 5 35 56 pm

- 	fix(address) state lines up with city
- 	feat(form) #333 - tooltips passed down to control

##### **Reviewers**
* @jgodi
* @more

### DIFF
--- a/src/elements/form/Control.ts
+++ b/src/elements/form/Control.ts
@@ -39,52 +39,52 @@ import { Helpers } from './../../utils/Helpers';
                         <div class="novo-control-input {{control.controlType}}" [ngSwitch]="control.controlType" [attr.data-automation-id]="control.key">
                             <!--Text-based Inputs-->
                             <!--TODO prefix/suffix on the control-->
-                            <div class="novo-control-input-container novo-control-input-with-label" *ngSwitchCase="'textbox'">
+                            <div class="novo-control-input-container novo-control-input-with-label" *ngSwitchCase="'textbox'" [tooltip]="control.tooltip" [tooltipPosition]="control.tooltipPosition">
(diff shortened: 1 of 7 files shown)


src/elements/form/Control.ts
```diff
@@ -39,52 +39,52 @@ import { Helpers } from './../../utils/Helpers';
                         <div class="novo-control-input {{control.controlType}}" [ngSwitch]="control.controlType" [attr.data-automation-id]="control.key">
                             <!--Text-based Inputs-->
                             <!--TODO prefix/suffix on the control-->
-                            <div class="novo-control-input-container novo-control-input-with-label" *ngSwitchCase="'textbox'">
+                            <div class="novo-control-input-container novo-control-input-with-label" *ngSwitchCase="'textbox'" [tooltip]="control.tooltip" [tooltipPosition]="control.tooltipPosition">
```
jgodi: Can you put this on the NovoFormControl (we need to do some re-naming) at https://github.com/amrutha-rajiv/novo-elements/blob/52e1d708872034b8d7550cf234402cd14bf9cb4a/src/elements/form/FormUtils.ts#L33? Then grab it via `form.controls[control.key].tooltip` and that. That way FieldInteractions can update the tooltip if needed.

jgodi: It will still go into the BaseControl and that, since we create those controls and make the NovoFormControl via them.

amrutha-rajiv: sure thing. Good call.

amrutha-rajiv:
@jgodi i have implemented those changes

PR #413, opened 2017-02-21
feat(chips): Added a preview on hover of the entity chip

ndickerson:
##### **Description**
Added a preview of the entity chip, which is exactly what was picked from the entity picker results.

* Pulled entity-picker-result out of entity-picker-results
* Added mouse over logic for the chip
* Added selection logic in the chips

##### **Reviewers**
* @jgodi
* @more


ndickerson:
chippreview

PR #416, opened 2017-02-23
fix(chips): Fixed the preview on hover of the entity chip when data is not present

ndickerson:
##### **Description**
Fixed the bug

##### **Reviewers**
* @jgodi
* @more


ndickerson:
chippreview

PR #422, opened 2017-03-01
updated novo tiles to have a disable option

rmcdole:
##### **Reviewers**
* @jgodi
* @more

### DIFF
--- a/src/elements/tiles/Tiles.scss
+++ b/src/elements/tiles/Tiles.scss
@@ -16,6 +16,10 @@ novo-tiles {
         cursor: pointer;
         align-items: center;
 
+        &.disabled {
+            opacity: .5;
+        }
+
         .tile {
             padding: 10px 15px;
             line-height: 1.5em;
--- a/src/elements/tiles/Tiles.scss
+++ b/src/elements/tiles/Tiles.scss
@@ -16,10 +16,6 @@ novo-tiles {
         cursor: pointer;
         align-items: center;
 
-        &.disabled {
-            opacity: .5;
-        }
-
         .tile {
             padding: 10px 15px;
             line-height: 1.5em;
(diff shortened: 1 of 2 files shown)


src/elements/tiles/Tiles.ts
```diff
@@ -43,6 +43,7 @@ export class NovoTilesElement implements ControlValueAccessor, OnInit {
     @Input() options: any;
     @Input() required: boolean;
     @Output() onChange: EventEmitter<any> = new EventEmitter();
+    @Input() enabled: boolean = true;
```
jgodi: Can you flip this, make the input `disabled`

src/elements/tiles/Tiles.scss
```diff
@@ -16,6 +16,10 @@ novo-tiles {
         cursor: pointer;
         align-items: center;
 
+        &.disabled {
+            opacity: .5;
```
jgodi: pointer-events: none; and then you dont need the check in the select

rmcdole: thats disabled clicking on it, which prevents our toast from happening.

jgodi: No, I mean if its disabled, then you probably shouldn't want to click on it at all. You have it:

```
        if (this.disabled){
            this.onChange.emit(item.value);
            this.onModelChange(item.value);
            return;
        }
```

which will stilll emit the change event for something that didn't get reflected in the UI.

if you just modify the CSS to have `pointer-events: none` then you won't need the logic above

rmcdole: ah, i see what you are saying. sounds like we should rename the variable instead then to make it more clear. ill think of something :thinking:

rmcdole: do you have any ideas?

rmcdole: Okay I added the toast logic to this class to hopefully make this more clear. If disabled, everything is disabled except toasts if toastOptions are passed in. Do you think the var name still needs to be changed?

PR #453, opened 2017-03-29
fix(FormUtils): Overrides for setValue calls are now outside of config

krsween:

Overrides for setValue calls are now outside of config

##### **Reviewers**
* @jgodi

### DIFF
--- a/src/elements/form/NovoFormControl.ts
+++ b/src/elements/form/NovoFormControl.ts
@@ -51,27 +51,44 @@ export class NovoFormControl extends FormControl {
      */
     public setRequired(isRequired: boolean): void {
         this.required = isRequired;
-        let validators: any = [...this.validators];
         // Update validators to have the required
         if (this.required && !this.hasRequiredValidator) {
+            let validators: any = [...this.validators];
             validators.push(Validators.required);
+            // TODO: duplicated below
+            this.setValidators(validators);
+            this.updateValueAndValidity();
+            this.hasRequiredValidator = this.required;
         } else if (!this.required && this.hasRequiredValidator) {
+            let validators: any = [...this.validators];
             validators = validators.filter(val => val !== Validators.required);
+            // TODO: duplicated above
+            this.setValidators(validators);
+            this.updateValueAndValidity();
+            this.hasRequiredValidator = this.required;
         }
-        this.setValidators(validators);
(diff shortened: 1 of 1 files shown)


src/elements/form/NovoFormControl.ts
```diff
@@ -51,27 +51,44 @@ export class NovoFormControl extends FormControl {
      */
     public setRequired(isRequired: boolean): void {
         this.required = isRequired;
-        let validators: any = [...this.validators];
         // Update validators to have the required
         if (this.required && !this.hasRequiredValidator) {
+            let validators: any = [...this.validators];
```
jgodi: Why are you doing this change? Looks like it was not duplicated before?

src/elements/form/NovoFormControl.ts
```diff
@@ -51,27 +51,44 @@ export class NovoFormControl extends FormControl {
...
+                        onlySelf?: boolean,
+                        emitEvent?: boolean,
+                        emitModelToViewChange?: boolean,
+                        emitViewToModelChange?: boolean
+                    } = {}) {
         this.markAsDirty();
         this.markAsTouched();
-        super.setValue(value, config);
+        // TODO: Should we set defaults on these?
+        super.setValue(value, { onlySelf, emitEvent, emitModelToViewChange, emitViewToModelChange });
```
jgodi: Should not set defaults, NG will call this internally in our forms in various places and it passes values to it. We only extend it here to always mark dirty/touched when setting a value.

src/elements/form/NovoFormControl.ts
krsween: It was in a previous commit. I DRY'd out the code and cause a bug because there's a case in between the if and the else that's needed for preventing fieldInteractions on NovoControls from being circular when they're set to fire onInit and onChange.

src/elements/form/NovoFormControl.ts
krsween: Seems odd to include all of this and then pass all of it through as undefined, but that's just my opinion.

src/elements/form/NovoFormControl.ts
jgodi: Oh, ya, Gotcha 👍

PR #482, opened 2017-06-06
feat(calendar): Adding ability to change and display current selected date

johnsully83:
##### **Description**
Added in ability to change and display current selected date in calendar module.

Added in CalendarDateChange component, cleaned up module a little, tweaked the demo a bit as well.

##### **Reviewers**
* @jgodi
* @more


src/elements/calendar/common/CalendarDateChange.ts
```diff
@@ -0,0 +1,76 @@
+import { Component, Input, Output, EventEmitter, Inject, LOCALE_ID } from '@angular/core';
+
+import * as addDays from 'date-fns/add_days';
```
jgodi: Use `import * as dateFns from 'date-fns';` and then `dateFns.addDays()`.

jgodi: Same for all the others too

PR #483, opened 2017-06-13
WIP: Converted the QuickNote to use CKEditor

ndickerson:
##### **Description**

Swapped out the plain text editor used in QuickNote for a right text editor

Replaced the old mechanism for using (sort of) rich text QuickNotes with CKEditor.

##### **Reviewers**
* @jgodi

### DIFF
--- a/src/elements/quick-note/QuickNote.ts
+++ b/src/elements/quick-note/QuickNote.ts
@@ -1,8 +1,6 @@
 // NG2
-import { Component, EventEmitter, forwardRef, ElementRef, ViewChild, ViewContainerRef, Input, Output, OnInit } from '@angular/core';
+import { Component, EventEmitter, forwardRef, ElementRef, ViewChild, ViewContainerRef, Input, Output, OnInit, AfterViewInit, OnDestroy, NgZone } from '@angular/core';
 import { NG_VALUE_ACCESSOR } from '@angular/forms';
-// Vendor
-import { Observable } from 'rxjs/Rx';
 // APP
 import { OutsideClick } from './../../utils/outside-click/OutsideClick';
 import { KeyCodes } from './../../utils/key-codes/KeyCodes';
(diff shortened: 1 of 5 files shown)


src/elements/quick-note/QuickNote.ts
```diff
@@ -16,62 +14,51 @@ const QUICK_NOTE_VALUE_ACCESSOR = {
...
-                      (focus)="onFocus($event)"
-                      (blur)="onTouched($event)"
-                      #textArea>
-            </textarea>
-            <div class="quick-note-overlay"
-                [innerHTML]="formattedNote"
-                [attr.placeholder]="placeholder"
-                 #quickNoteOverlay>
-            </div>
+            <textarea #host></textarea>
```
jgodi: Can this just use the CKEditor component that we already have? Would clean up some of the class?

jgodi:
Can we get focus into the CKeditor when they select an item? 

zxvj40ekpa

jgodi:
The CustomResultsTemplate demo is broken too, you are probably aware just documenting it.

`No component factory found for CustomQuickNoteResults`

src/elements/quick-note/QuickNote.ts
ndickerson: Will look into this.

ndickerson: Will look into this.

ndickerson: Will look into this.

ndickerson:
Stripping links has been removed in my corresponding novo branch. 

CustomQuickNoteResults is broken on master

Will test the back spacing more thoroughly. 

We probably don't need cleanup of removed references because while that behavior is possible in novo elements today, we don't remove references in novo. They stick around even after the mention is gone.

jgodi:
Ah, it does. We really need E2E tests in here to catch these things.

I think the clean up of references is important, since this is an open source component and should work even though that Novo doesn't use it in the same way. Or this element should be removed from open source. Otherwise they can backspace just a few characters, keep the link in there and the reference would persist.

PR #486, opened 2017-06-16
Converted the QuickNote to use CKEditor

ndickerson:
##### **Description**

Swapped out the plain text editor used in QuickNote for a right text editor

Replaced the old mechanism for using (sort of) rich text QuickNotes with CKEditor.

##### **Reviewers**
* @jgodi


jgodi:
The only thing that I notice is that the references aren't cleaned up, so if I do #first and then delete it, the reference is there then if I do #first again I don't get a link.

Other than that it looks great.

ndickerson:
Added the dynamic removal of references when their rendered text has been removed from the note

PR #490, opened 2017-06-22
bug(Editable tables): Percentage cells do not render in the editable form.

MagicAardvark:
##### **Description**
Table cells have a percentage renderer that multiplies the current value by 100 (PercentageCell.ts) however when it is toggled over to edit mode the percentages show in their original 0.2 (20%) format. this is because the percentage renderer lives in bh-elements.

now logic would say we should move the percentage renderer to novo-elemnets. however this produces another problem because 0.2 then becomes 2000% because it is multiplied for both the form and the table.

Added some specialty renderers which would perform this calculations based on dataSpecialization. these are added when the form is toggled to edit mode only, and removed when the form leaves edit mode.

##### **Reviewers**
* @jgodi
* @more

### DIFF
--- a/src/elements/table/Table.spec.ts
+++ b/src/elements/table/Table.spec.ts
@@ -421,7 +421,8 @@ describe('Elements: NovoTableElement', () => {
     describe('Method: setTableEdit()', () => {
         beforeEach(() => {
             component._dataProvider = {
-                edit: () => {}
+                edit: () => { },
+                refresh: () => { }
             };
             component.columns = [
                 {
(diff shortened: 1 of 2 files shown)


bvkimball:
I don't know if i agree that Everyone wants to edit percentages as percentages. I feel like this is a less common paradigm.  I know our forms may work this way but i am not so sure this is right.

MagicAardvark:
I agree it may not be the best long term. Unfortunately given the current architecture I have been unable to solve it in a better manner. I feel we will have to rework edit tables quite a bit in order to cover a variety of cases as we use them more.

jgodi:
I think we need to redo edit tables anyway, break out from the normal table and then move other stuff into shared components.

Issue #436, opened 2017-03-16
[Editable Table] - Changed visible row data does not revert when cancel is selected


MagicAardvark:
_**Note: for support questions, please use one of these channels:**_ [Chat: BullhornOSS Gitter](https://gitter.im/bullhorn/Open-Source)

##### **I'm submitting a BUG REPORT | FEATURE REQUEST | QUESTION]**
BUG REPORT

##### **What is the current behavior?**
Cancel does not correctly revert the data

##### **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Currently when using the editable table a user: 
1. enters Edit mode
2. edits one or more fields.
3. Hits cancel

##### **What is the expected behavior?**
The tableForm data reflects the cancel correctly but the data in the table itself shows the changed data. This is due to the table-cell component having an outdated form reference.

##### **Please tell us about your environment:**

- Angular version: 2.0.0-RC.X
- Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

##### **Other information**
_(e.g. detailed explanation, screenshots, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)_

jgodi:
@MagicAardvark Didn't you fix this?

PR #491, opened 2017-06-23
Editable Table Fixes

jgodi:
Fixes https://github.com/bullhorn/novo-elements/issues/436


src/services/data-provider/ArrayCollection.ts
```diff
@@ -68,15 +68,15 @@ export class ArrayCollection<T> implements Collection<T> {
      */
     edit() {
         this.isEditing = true;
-        this.editData = this.source.slice();
+        this.editData = Helpers.deepClone(this.source);
```
MagicAardvark: are we renaming deepClone to copy and use it as utility to  call deepClone?

PR #493, opened 2017-06-27
feat(Radio): Adding ability to have icon only button radio groups

MagicAardvark:
##### **Reviewers**
* @jgodi
* @more


jgodi:
@MagicAardvark Build failing

PR #494, opened 2017-06-28
BH-51802 call hideResults when we remove the search symbol

nathan-wilson-bh:
##### **Reviewers**
* @jgodi
* @more

### DIFF
--- a/src/elements/quick-note/QuickNote.ts
+++ b/src/elements/quick-note/QuickNote.ts
@@ -394,6 +394,11 @@ export class QuickNoteElement extends OutsideClick implements OnInit, OnDestroy,
         let word = this.getWordAtCursor().trim();
         if (this.isTagging) {
             let symbol = this.config.triggers[this.taggingMode];
+            if (word.indexOf(symbol) === -1) {
+                this.zone.run(() => {
+                    this.hideResults();
+                });
+            }
             word = word.slice(word.indexOf(symbol) + symbol.length);
         }
         return word;
--- a/src/elements/quick-note/QuickNote.ts
+++ b/src/elements/quick-note/QuickNote.ts
@@ -394,10 +394,9 @@ export class QuickNoteElement extends OutsideClick implements OnInit, OnDestroy,
         let word = this.getWordAtCursor().trim();
         if (this.isTagging) {
             let symbol = this.config.triggers[this.taggingMode];
-            if (word.indexOf(symbol) === -1) {
-                this.zone.run(() => {
-                    this.hideResults();
-                });
+            if (!word.includes(symbol)) {
+                this.hideResults();
+                return '';
             }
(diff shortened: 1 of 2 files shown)


src/elements/quick-note/QuickNote.ts
```diff
@@ -394,6 +394,11 @@ export class QuickNoteElement extends OutsideClick implements OnInit, OnDestroy,
         let word = this.getWordAtCursor().trim();
         if (this.isTagging) {
             let symbol = this.config.triggers[this.taggingMode];
+            if (word.indexOf(symbol) === -1) {
+                this.zone.run(() => {
```
ndickerson: We don't need the zone.run here - at this point, we are already in the Angular context.

src/elements/quick-note/QuickNote.ts
```diff
@@ -394,6 +394,11 @@ export class QuickNoteElement extends OutsideClick implements OnInit, OnDestroy,
         let word = this.getWordAtCursor().trim();
         if (this.isTagging) {
             let symbol = this.config.triggers[this.taggingMode];
+            if (word.indexOf(symbol) === -1) {
```
ndickerson: Can we do a `!word.includes(...)` instead?

src/elements/quick-note/QuickNote.ts
nathan-wilson-bh: I took out the zone.run and called hideResults directly and the component never actually gets destroyed

ndickerson: Ok. Can we add a unit test case for this?

nathan-wilson-bh: Yeah I planned on doing that. Just wanted to get the implementation out there for review first

nathan-wilson-bh: So turns out I didn't need to run it in the angular zone and it was somehow just masking a bigger issue. Now it just returns empty string after it hides and works without running it in the angular zone

src/elements/quick-note/QuickNote.ts
nathan-wilson-bh: Done.

src/elements/quick-note/QuickNote.ts
```diff
@@ -394,6 +394,10 @@ export class QuickNoteElement extends OutsideClick implements OnInit, OnDestroy,
         let word = this.getWordAtCursor().trim();
         if (this.isTagging) {
             let symbol = this.config.triggers[this.taggingMode];
+            if (!word.includes(symbol)) {
```
jgodi: Spec?

nathan-wilson-bh: Coming in the next commit

nathan-wilson-bh: Done

PR #503, opened 2017-07-10
style(Dropdown): Change width for selected dropdown values

akurtovic:
##### **Description**
Minor style change to allow dropdown elements to take up the width of the form.

##### **Reviewers**
* @jgodi
* @more


jgodi:
I don't think we want this fix here in NovoElements. I think the fix should just be in Novo code since we only would want this in our cards and this will change it everywhere that novo-dropdowns are used.

PR #504, opened 2017-07-11
fix(picker): Fixing picker to accept right click paste values

monroepe:
##### **Description**
This fix allows users to right click and paste values into pickers.

I added an observer on paste events for pickers so that pickers can handle right click pasted values.

##### **Reviewers**
* @jgodi
* @more


jgodi:
I will check this out and merge it tonight :)

PR #507, opened 2017-07-18
fix(EntityPickerResults): EntityPicker was using bad value for date f…

bvkimball:
Error Message: 
novo-elements/src/elements/picker/extras/entity-picker-results/EntityPickerResults.ts(124,64): error TS2559: Type '"L"' has no properties in common with type 'DateTimeFormatOptions'.

I am using the Intl Date format options now.  I believe this was a left over bug from our old date formatting code.

##### **Reviewers**
* @jgodi
* @more


jgodi:
#### Overall Coverage Change: -0.37%
|File|Change|
|---|---|
|EntityPickerResults.ts|No Change|

PR #508, opened 2017-07-18
fix(calendar): Fixing some styles around calendar events and having m…

johnsully83:
…ultiple on the same day

##### **Description**
Fixing some issues around calendar, mostly around events displaying.

Formatted some SCSS that we had previously copied from an external source; similarly made some edits to that CSS to help with z-index on tooltips, allowing display of multiple events on the same day in the week view, and modifying the height of an event dependent on it's duration.

##### **Reviewers**
* @jgodi
* @more


jgodi:
Build fails, please fix and I will merge

PR #535, opened 2017-08-21
fix(select): Fixing select searching and selecting

monroepe:
## **Description**

Searching for values in select drop downs is currently broken and this pull request fixes the issue. Basically if you open a select element (where the options are just strings as seen in the demo) and start typing you are unable to select any of the values. The fix is that we check if the options are not objects and use the formattedOptions instead.

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**

### DIFF
--- a/src/elements/select/Select.ts
+++ b/src/elements/select/Select.ts
@@ -176,11 +176,7 @@ export class NovoSelectElement extends OutsideClick implements OnInit, OnChanges
                     list.scrollTop = item.offsetTop;
                     let listItems = Array.from(list.querySelectorAll('li')).map((element: any) => element.getAttribute('data-automation-value')).filter(value => value);
                     this.selectedIndex = listItems.indexOf(item.getAttribute('data-automation-value'));
-                    if (this.options && this.options.length && typeof this.options[0] !== 'object') {
-                        this.select(this.filteredOptions[this.selectedIndex], this.selectedIndex);
-                    } else {
-                        this.select(this.options[this.selectedIndex], this.selectedIndex);
-                    }
+                    this.select(this.filteredOptions[this.selectedIndex], this.selectedIndex);
                 }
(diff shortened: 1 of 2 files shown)


src/elements/select/Select.ts
```diff
@@ -174,9 +174,13 @@ export class NovoSelectElement extends OutsideClick implements OnInit, OnChanges
                 let item = element.querySelector(`[data-automation-value^="${this.filterTerm}" i]`);
                 if (item) {
                     list.scrollTop = item.offsetTop;
-                    let listItems = Array.from(list.querySelectorAll('li')).map((element: any) => element.getAttribute('data-automation-value'));
+                    let listItems = Array.from(list.querySelectorAll('li')).map((element: any) => element.getAttribute('data-automation-value')).filter(value => value);
                     this.selectedIndex = listItems.indexOf(item.getAttribute('data-automation-value'));
-                    this.select(this.options[this.selectedIndex], this.selectedIndex);
+                    if (this.options && this.options.length && typeof this.options[0] !== 'object') {
```
jgodi: Shouldn't this always be `filteredOptions`?

monroepe: Yeah I originally had it that way always using filteredOptions and was going to ask you if that was ok, so I will switch it to always use filteredOptions.

ndickerson:
The changes look good for the up/down arrow keys, but we should address all the areas of the code where we were using the non-filteredOptions.

jgodi:
#### Overall Coverage Change: -0.91%
|File|Change|
|---|---|
|DatePicker.ts|No Change|
|FileInput.ts|-0.31%|
|Select.ts|-0.89%|
|Table.ts|-0.29%|

jgodi:
#### Overall Coverage Change: -0.91%
|File|Change|
|---|---|
|DatePicker.ts|No Change|
|FileInput.ts|-0.31%|
|Select.ts|-0.89%|
|Table.ts|-0.29%|

PR #542, opened 2017-09-06
Added min-width to address control to make flex behave in Firefox

nathan-wilson-bh:
## **Description**

Added min-width to address control inputs to make flex behave in Firefox

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**

### DIFF
--- a/src/elements/form/extras/address/Address.scss
+++ b/src/elements/form/extras/address/Address.scss
@@ -3,6 +3,7 @@ novo-address {
     flex-flow: row wrap;
 
     input {
+        min-width: 10px;
     }
 
     .street-address {


jgodi:
|File (0)|Change (:white_check_mark: No Change)|
|---|---|

src/elements/form/extras/address/Address.scss
```diff
@@ -3,6 +3,7 @@ novo-address {
     flex-flow: row wrap;
 
     input {
+        min-width: 10px;
```
jgodi: Seems like a hack? Is the overall way we are doing the flex css around the address incorrect?

PR #547, opened 2017-09-13
(feat) Field Interaction async loading animations, add/remove fields and open lists

akurtovic:
## **Description**
Work on FieldInteractionApi and AppBridge

- Changed AppBridge to create method for opening lists
- Changed FieldInteractionAPI to allow adding/removing fields on events other than onInit
- Added loading animation for async field interactions in Control.ts
- Styling changes to Form to allow for loading animation

#### **Verify that...**

- [ X ] Any related demos where added and `npm start` still works
- [ X ] New demos work in `Safari`, `Chrome` and `Firefox`
- [ X ] `npm run lint` passes
- [ ] `npm test` passes and code coverage is increased (no unit tests added)
- [ X ] `npm run compile` still works

##### **Screenshots**
bluespinner

### DIFF
--- a/src/elements/form/FieldInteractionApi.ts
+++ b/src/elements/form/FieldInteractionApi.ts
@@ -438,6 +438,7 @@ export class FieldInteractionApi {
         let control = this.getControl(key);
         if (control) {
             if (loading) {
+                this.form.controls[key].fieldInteractionloading = true;
                 control.setErrors({ 'loading': true });
                 // History
                 clearTimeout(this.asyncBlockTimeout);
@@ -447,6 +448,7 @@ export class FieldInteractionApi {
                     this.setProperty(key, '_displayedAsyncFailure', true);
                 }, 10000);
             } else {
+                this.form.controls[key].fieldInteractionloading = false;
                 clearTimeout(this.asyncBlockTimeout);
                 control.setErrors({ 'loading': null });
                 control.updateValueAndValidity({ emitEvent: false });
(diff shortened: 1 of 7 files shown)


src/elements/form/FieldInteractionApi.ts
```diff
@@ -437,7 +437,8 @@ export class FieldInteractionApi {
     public setLoading(key: string, loading: boolean) {
         let control = this.getControl(key);
         if (control) {
-            if (loading) {
+            if (loading && control.value) {
```
jgodi: Need to check the value here? If the user nulls our the field we still probably want to perform the async too

src/elements/form/Form.module.ts
```diff
@@ -26,6 +26,8 @@ import { NovoDragulaModule } from './../dragula/Dragula.module';
 import { NovoTipWellModule } from './../tip-well/TipWell.module';
 import { NovoModalModule } from './../modal/Modal.module';
 import { ControlConfirmModal } from './ControlConfirmModal';
+import { NovoLoadingModule } from '../loading/Loading.module';
```
jgodi: Don't need the loading module anymore?

src/elements/form/FieldInteractionApi.ts
akurtovic: I added this because the animation was going when you wipe out the field. It looked weird to me. I can take out the check if that's the desired behavior (including the loading animation) or I can add another check in the ngIf in template to check for value there and only show the loading animation if the field has a value.

src/elements/form/Form.module.ts
akurtovic: Yup. Good catch.

src/elements/form/FieldInteractionApi.ts
akurtovic: Actually, now that I think about it, I probably don't need this. An async picker field would still need to load when you wipe out the value to get all the options, etc.

jgodi:
|File (2)|Change (:white_check_mark: +0.61%)|
|---|---|
|Select.module.ts|:white_check_mark: +100%|
|Select.ts|:white_check_mark: +3.54%|

PR #561, opened 2017-10-04
fix: event was no longer used but still lingered

bvkimball:
## **Description**

event was left behind and some browser thought it was window.event so it worked fine.


jgodi:
|File (1)|Change (:white_check_mark: No Change)|
|---|---|
|Overlay.ts|:white_check_mark: No Change|

PR #570, opened 2017-11-08
added the encrypted icon to the control form when the field is a customEncrypted field

rmcdole:
encryptedfields
## **Description**

added the encrypted icon to the control form when the field is a customEncrypted field

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**

### DIFF
--- a/src/elements/form/Control.ts
+++ b/src/elements/form/Control.ts
@@ -99,12 +99,14 @@ export class NovoCustomControlContainerElement {
     selector: 'novo-control',
     template: `
         <div class="novo-control-container" [formGroup]="form" [hidden]="form.controls[control.key].hidden || form.controls[control.key].type === 'hidden' || form.controls[control.key].controlType === 'hidden'">
-            <!--Encrypted Field-->
-            <i [hidden]="!isFieldEncrypted(control)"
-               class="bhi-lock">
-            </i>
             <!--Label (for horizontal)-->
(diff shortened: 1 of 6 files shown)


jgodi:
|File (1)|Change (:white_check_mark: Pass)|
|---|---|
|Control.ts|:white_check_mark: +0.23%|

src/elements/form/Control.ts
```diff
@@ -99,6 +99,10 @@ export class NovoCustomControlContainerElement {
     selector: 'novo-control',
     template: `
         <div class="novo-control-container" [formGroup]="form" [hidden]="form.controls[control.key].hidden || form.controls[control.key].type === 'hidden' || form.controls[control.key].controlType === 'hidden'">
+            <!--Encrypted Field-->
+            <i [hidden]="!isFieldEncrypted(control)"
```
jgodi: Don't want to call a function inside of interpolation. Angular won't know how to handle change detection for this and it will slow down forms. 

How are you marking the field as encrypted? Just putting the words `customEncrypted` on the key? Why not add a new property to the BaseControl, propagate it onto the NovoFormControl and then you can just `form.controls[control.key].encrypted` here?

jgodi: Also, does the field need to render differently if it's encrypted? Should it be disabled or obfuscated like a password field?

rmcdole: Good call, working on making those changes. As for the purpose, just needed to make an indicator that the data is encrypted in the db. What you are talking about is confidential fields, which we are implementing in another epic.

escarre:
Updated @jgodi

PR #642, opened 2018-01-31
F/sms

jaclynjessup:
## **Description**

- Removed phone & email from value
- Value with icons right now takes multiple icons

#### **Verify that...**

- [x] `npm run lint` passes


amrutha-rajiv:
can you remove novo-value email as well with this PR?

src/platform/elements/value/Value.ts
```diff
@@ -131,11 +100,16 @@ export class NovoValueElement implements OnInit, OnChanges {
...
 
-    public get iconClass(): string {
-        if (this.meta && this.meta.icon) {
-            return `bhi-${this.meta.icon} actions`;
+    iconClass(icon): string {
+        let iconClass = '';
+        if (icon && icon.iconCls) {
+            iconClass = `bhi-${icon.iconCls} actions`;
+            if (icon.onIconClick) {
+                iconClass = `${iconClass} clickable`
```
amrutha-rajiv: can you add styles for the actions class to hide the cursor (pointer) on hover and only show the pointer cursor when clickable is active?

jaclynjessup: yes will do

PR #660, opened 2018-02-14
BH-54679 Adding entity list to value renderer again

DrewDudley:
## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**


src/platform/elements/value/EntityList.ts
```diff
@@ -0,0 +1,74 @@
...
+        return this.ENTITY_SHORT_NAMES[entity.personSubtype];
+    }
+
+    openLink(entity: any): void {
+        entity.openLink(entity);
+    }
+
+    isLinkable(entity: any): boolean {
+        let type: any = entity.personSubtype || this.baseEntity;
+        return ['Candidate', 'ClientContact', 'Lead', 'Opportunity', 'Placement', 'JobOrder', 'ClientCorporation', 'Appointment', 'Task'].indexOf(type) >= 0;
```
amrutha-rajiv: `isLinkable` should return `true` only if  `entity.openLink` is present

src/platform/elements/value/Value.ts
```diff
@@ -114,4 +117,11 @@ export class NovoValueElement implements OnInit, OnChanges {
         let isURL: any = Helpers.isString(data) && regex.exec(data.trim());
         return (linkFields.indexOf(field.name) > -1) || !!isURL || field.type === NOVO_VALUE_TYPE.LINK;
     }
+
+    isEntityList(type: string, associatedEntity: { entity: string, entityMetaUrl: string, label: string, fields: any[] }): boolean {
+        if (associatedEntity && associatedEntity.entity === 'Note') {
+            return false;
+        }
+        return type === 'TO_MANY';
+    }
```
amrutha-rajiv: i think it should be fine to just default to `EntityList` for all to_many OR have the user of `novo-value` set valueType or something to `entity-list` and only display as entity-list if the valueType is set to that value

src/platform/elements/value/EntityList.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    openLink(entity: any): void {
+        entity.openLink(entity);
+    }
+
+    isLinkable(entity: any): boolean {
+        let type: any = entity.personSubtype || this.baseEntity;
+        return ['Candidate', 'ClientContact', 'Lead', 'Opportunity', 'Placement', 'JobOrder', 'ClientCorporation', 'Appointment', 'Task'].indexOf(type) >= 0 && entity.openLink;
+    }
```
amrutha-rajiv: i think this has to be be just if `entity.openLink`

PR #672, opened 2018-03-09
feat(value) displaying html fields for mobile

tgadre:
## **Description**

Added a case for html/text area fields display. The css class is used in mobile to add custom styles
Also added a flag to strip out the html
screen shot 2018-03-09 at 3 53 53 pm

Novo will not be affected and is being tested

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**

### DIFF
--- a/src/platform/elements/value/Value.spec.ts
+++ b/src/platform/elements/value/Value.spec.ts
@@ -184,12 +184,19 @@ describe('Elements: NovoValueElement', () => {
             component.ngOnChanges();
             expect(component.type).toEqual(NOVO_VALUE_TYPE.LINK);
         });
-        it('should set type to html for a html fields', () => {
+        it('should set type to html for a large fields', () => {
             component.meta.name = 'companyDescription';
             component.meta.dataSpecialization = 'HTML';
             component.data = '';
             component.ngOnChanges();
             expect(component.type).toEqual(NOVO_VALUE_TYPE.HTML);
+        });
+        it('should strip html tags in large fields', () => {
+            component.meta.name = 'companyDescription';
+            component.meta.dataSpecialization = 'HTML';
+            component.data = '<span style="color:#8e44ad">test</span>';
+            component.ngOnChanges();
+            expect(component.data).toEqual('test');
         });
     });
     describe('Function: isLinkField', () => {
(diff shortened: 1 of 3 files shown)


jgodi:
|File (1)|Change (:bangbang: -25.75%)|
|---|---|
|Value.ts|:white_check_mark: +0.67%|

src/platform/elements/value/Value.spec.ts
```diff
@@ -5,7 +5,7 @@ import { NovoValueElement, NOVO_VALUE_THEME, NOVO_VALUE_TYPE } from './Value';
 // import { NovoLabelService } from '../../services/novo-label-service';
 import { NovoValueModule } from './Value.module';
 // TODO fix specs
-describe('Elements: NovoValueElement', () => {
+fdescribe('Elements: NovoValueElement', () => {
```
amrutha-rajiv: fdescribe

src/platform/elements/value/Value.ts
```diff
@@ -96,6 +100,11 @@ export class NovoValueElement implements OnInit, OnChanges {
             }
         } else if (this.isEntityList(this.meta.type)) {
             this.type = NOVO_VALUE_TYPE.ENTITY_LIST;
+        } else if (this.isHTMLField(this.meta)) {
+            this.type = NOVO_VALUE_TYPE.HTML;
+            if (this.meta.stripHTML) {
+                this.data = this.data.replace(/<(.|\n)+?>/gi, '');
```
amrutha-rajiv: this can probably be done by the render pipe

src/platform/elements/value/Value.ts
```diff
@@ -16,6 +16,10 @@ export enum NOVO_VALUE_THEME { DEFAULT, MOBILE };
                 <novo-entity-list *ngSwitchCase="NOVO_VALUE_TYPE.ENTITY_LIST" [data]='data' [meta]="meta"></novo-entity-list>
             </div>
 
+            <div *ngSwitchCase="NOVO_VALUE_TYPE.HTML" class="value-outer html-field">
+                <label>{{ meta.label }}</label>
```
amrutha-rajiv: if the class is the only thing that is different for `HTML` then we don't need a switchcase within the template - we can use some kind of variable to retrieve class value and use it in the default case

src/platform/elements/value/Value.spec.ts
tgadre: done

src/platform/elements/value/Value.ts
tgadre: done

PR #698, opened 2018-04-14
Add support into datepicker for looser date formats and different locale options

charlesabarnes:
## **Description**

I added an option for moment and I pass in the format option through a couple of places so it actually get used in the component.   

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**

image


package.json
```diff
@@ -52,13 +52,14 @@
...
     "@angular/cdk": "5.2.4",
-    "@types/dragula": "2.1.29",
     "@bullhorn/dragula": "1.0.1",
+    "@types/dragula": "2.1.29",
     "ace-builds": "1.3.1",
     "angular2-text-mask": "8.0.4",
     "brace": "0.11.1",
     "date-fns": "1.29.0",
     "hint.css": "2.5.0",
+    "moment": "^2.22.0",
```
jgodi: We do not want moment as a dependency in NovoElements. It is too bloated. We had everything use moment in that past and updated to use `date-fns`.

jgodi: We should find a solution using date-fns without another dep.

jgodi: https://date-fns.org/docs/format

Can still have custom formatting, if pass custom format, disable the mask

charlesabarnes: Good call, I got tunnel vision. I will correct it.

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -45,20 +46,28 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
   @Input() placeholder: string;
   @Input() maskOptions: any;
   @Input() format: string;
+  @Input() specialFormat: boolean;
```
jgodi: All the code looks solid. Not loving the name `specialFormat`. Can you think of a better name that is more descriptive?

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -45,20 +46,28 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
   @Input() placeholder: string;
   @Input() maskOptions: any;
   @Input() format: string;
+  @Input() userDefinedFormat: boolean;
```
jgodi: Sorry, I just had a thought, could you still use format but do a regex on the `this.format` to check if its `dd/dd/dddd` to use the mask vs not?

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -45,20 +46,28 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
...
   ngOnInit() {
-    this.maskOptions = this.maskOptions || {
-      mask: [/\d/, /\d/, '/', /\d/, /\d/, '/', /\d/, /\d/, /\d/, /\d/],
-      pipe: createAutoCorrectedDatePipe(this.format || this.labels.dateFormat.toLowerCase()),
-      keepCharPositions: false,
-      guide: true,
-    };
+    if(!this.userDefinedFormat) {
+      this.maskOptions = this.maskOptions || {
+        mask: [/\d/, /\d/, '/', /\d/, /\d/, '/', /\d/, /\d/, /\d/, /\d/],
```
jgodi: Need to keep the pipe?

charlesabarnes: forgot to remove.  Updated Thanks!

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -4,7 +4,7 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { TAB, ENTER, ESCAPE } from '@angular/cdk/keycodes';
 // Vendor
 import { TextMaskModule } from 'angular2-text-mask';
-import createAutoCorrectedDatePipe from 'text-mask-addons/dist/createAutoCorrectedDatePipe';
```
jgodi: Don't need this anymore? Does this break what Novo is expecting?

charlesabarnes: I confused that with another change I made earlier. I added it back in

PR #706, opened 2018-04-25
Date format Part2

charlesabarnes:
## **Description**

I added an option for moment and I pass in the format option through a couple of places so it actually get used in the component.   

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**

image


src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -45,20 +47,30 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
   @Input() placeholder: string;
   @Input() maskOptions: any;
   @Input() format: string;
+  @Input() textMaskEnabled: boolean;
```
jgodi: Can default to true, then you don't need to check for undefined below

jgodi: nvm, I see how you are doing it, maybe flip it? `disabledTextMask: boolean = false` then down below itll be `!!field.disableTextMask`

jgodi:
|File (2)|Change (:bangbang: -0.26%)|
|---|---|
|Address.ts|:white_check_mark: +0.00%|
|FormUtils.ts|:white_check_mark: +0.00%|

src/platform/elements/date-picker/DatePickerInput.ts
charlesabarnes: You were right.  I wanted it to just default to true. I just set the default.  It works just fine.

PR #719, opened 2018-05-14
Add Emitter for NovoFormControl Changes

charlesabarnes:
## **Description**

Emit if there are changes to the NovoFormControl via Field Interactions

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**

### DIFF
--- a/src/platform/elements/form/FieldInteractionApi.ts
+++ b/src/platform/elements/form/FieldInteractionApi.ts
@@ -13,6 +13,7 @@ import { ControlConfirmModal, ControlPromptModal } from './FieldInteractionModal
 import { Helpers } from '../../utils/Helpers';
 import { AppBridge } from '../../utils/app-bridge/AppBridge';
 import { NovoLabelService } from '../../services/novo-label-service';
+import { IFieldInteractionEvent} from './FormInterfaces';
 
 @Injectable()
 export class FieldInteractionApi {
@@ -153,7 +154,7 @@ export class FieldInteractionApi {
         let control = this.getControl(key);
         if (control) {
             control.setValue(value, options);
-            this.form.novoConfigChange.emit({value: value});
+            this.form.triggerEvent({controlKey: key, prop: 'value', value: value });
         }
     }
 
@@ -166,6 +167,7 @@ export class FieldInteractionApi {
         let control = this.getControl(key);
         if (control) {
             control.setValue(value, options);
+            this.form.triggerEvent({controlKey: key, prop: 'value', value: value });
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/platform/elements/form/NovoFormControl.ts
```diff
@@ -195,7 +206,9 @@ export class NovoFormControl extends FormControl {
     }
 }
 
+
 export class NovoFormGroup extends FormGroup {
+    @Output() public novoConfigChange: EventEmitter<any> = new EventEmitter();
```
jgodi: No output here, and make a specific interface for the data model. Make sure to export the interface inside of `index.ts`

src/platform/elements/form/NovoFormControl.ts
```diff
@@ -184,6 +186,15 @@ export class NovoFormControl extends FormControl {
         }
     }
 
+    /**
+     * sets the `tooltip` property on the control
+     * @name setTooltip
+     * @param tooltip 
+     */
+    public setTooltip(tooltip: string): void {
```
jgodi: Don't need this anymore?

src/platform/elements/form/FieldInteractionApi.ts
```diff
@@ -311,7 +318,8 @@ export class FieldInteractionApi {
     public setTooltip(key: string, tooltip: string): void {
         let control = this.getControl(key);
         if (control) {
-            control.tooltip = tooltip;
+            control.setTooltip(tooltip);
+            this.form.novoConfigChange.emit({tooltip: tooltip});
```
jgodi: Make a private method to do this, we will need some defensive programming around this.

```
private triggerEvent(event: IFieldInteractionEvent) {
  if (this.form && this.form.fieldInteractionEvents) {
    this.form.fieldInteractionEvents.emit(event);
  }
}
```

Rename the `novoConfigChange` to `fieldInteractionEvents` to be a bit more generic. Then the data model should be like:

```
{
   controlKey: string;
   prop: string;
   value: any;
}
```

then in  each method (not just the ones you need) after the FI do the `this.triggerEvent({ controlKey: key, prop: 'tooltip', value: 'New Tooltip' });`

src/platform/elements/form/NovoFormControl.ts
```diff
@@ -1,9 +1,11 @@
 // NG2
 import { FormControl, Validators, FormGroup } from '@angular/forms';
-import { EventEmitter } from '@angular/core';
+import { EventEmitter, Output } from '@angular/core';
```
jgodi: output, observable unused

PR #725, opened 2018-05-21
Allowing for header icon/theme to be changed

ndickerson:
## **Description**

Allows the header icon or theme to be changed without having to destroy/recreate the element.

#### **Verify that...**

- [ ] `npm run lint` passes

##### **Screenshots**


src/platform/elements/header/Header.ts
```diff
@@ -55,20 +55,23 @@ export class NovoUtilActionComponent {
...
     @HostBinding('class.condensed') @Input() public condensed: boolean = false;
 
     public inverse: string = 'inverse';
     public iconClass: string;
 
     public ngOnInit(): void {
+        this.ngOnChanges();
+    }
+
+    public ngOnChanges(): void {
```
jgodi: Actually can you just remove ngOnInit and onChanges? Use setters/getters instead. Look at button for what it  looks like

ndickerson: Done

PR #734, opened 2018-06-05
fix(NovoAutoSize): Fix shrinking behavior of autosizing textareas

tb9-jeremy:
## **Description**

The autosizing textarea was shrinking by one pixel on every input event, depending on content/initial height of the textarea. The method of resizing is now `resize to smallest valid input height to calculate the appropriate scrollHeight`. The heights previously set in the directive have also been moved to min/max heights in CSS. Also removed some padding to better replicate the look of a one line text input.

#### **Verify that...**

- [x] `npm run lint` passes

The `compile` script does not seem to exist

##### **Screenshots**

Old behavior:
autosize-old

New behavior:
autosize-new

### DIFF
--- a/src/platform/elements/form/Control.spec.ts
+++ b/src/platform/elements/form/Control.spec.ts
@@ -1,0 +1,65 @@
+// NG2
+import { Component } from '@angular/core';
+import { TestBed, async } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+// App
+import { NovoAutoSize } from './Control';
+
+@Component({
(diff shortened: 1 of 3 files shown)


jgodi:
Can you verify the functionality when there is an initial value that is both (1) short and (2) very long/multiple lines.

tb9-jeremy:
Excuse the 3ish second delay when these gifs start, they are showing page reload. I can add specific test cases for these if you'd like.

Initialized with single line:
single line

Initialized over max height:
full

PR #736, opened 2018-06-05
Updates to allow override templates to all types of form controls

amrutha-rajiv:
## **Description**
- ControlTemplates house templates for all controls
- templates can be overriden via templates added to form
- custom controls can be added using ng-templates as well
#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**


src/platform/elements/form/DynamicForm.ts
```diff
@@ -80,129 +55,139 @@ export class NovoFieldsetElement {
...
+        this.form.fieldsets = [...this.fieldsets];
+    }
 
-        // Hide required fields that have been successfully filled out
-        if (hideRequiredWithValue && !Helpers.isBlank(this.form.value[control.key]) && (!control.isEmpty || control.isEmpty && control.isEmpty(this.form.controls[control.key]))) {
-          this.form.controls[control.key].hidden = true;
+    ngAfterContentInit() {
+        if (this.customTemplates && this.customTemplates.length) {
+            this.customTemplates.forEach((template: any) => {
+                this.templates.addCustom(template.type, template.template);
```
bvkimball: Wouldn't this make defining the component on one form add it to all forms?

I would think we would want to add them independently?

amrutha-rajiv: Yea, I was thinking the same. I think you should be able to do both. App level overrides/custom templates and form level overrides. What do you think?

jgodi:
Can you resolve the spacing issues? There is an editor-config file and there are extensions for IDEs to read that in and update he spacing settings based on that. 

Will help moving from project to project where the spacing could be different.

src/app/pages/elements/form/FormDemo.ts
```diff
@@ -387,7 +387,7 @@ export class FormDemoComponent {
       { token: 'TOKEN', military: true },
       {
         customfield: {
-          customControl: CustomDemoComponent,
+          customControl: 'custom-demo-component',
```
jgodi: Can this be renamed to `template` just to align with tables? They pass `template` to the column config to specify a custom cell.

src/app/pages/elements/form/templates/VerticalDynamicForm.html
```diff
@@ -1,6 +1,10 @@
 <button theme="secondary" (click)="nomyform.showAllFields()">Show All Fields</button>
 <button theme="secondary" (click)="nomyform.showOnlyRequired()">Show Required Fields</button>
-<novo-dynamic-form layout="vertical" class="dynamic" [controls]="dynamicVertical" [(form)]="dynamicVerticalForm" #nomyform></novo-dynamic-form>
+<novo-dynamic-form layout="vertical" class="dynamic" [controls]="dynamicVertical" [(form)]="dynamicVerticalForm" #nomyform>
+  <ng-template novoTemplate="custom-demo-component" type="custom-demo-component" let-control let-form="form">
```
jgodi: What does the `type` do? That is already passed in the `novoTemplate` right? The value is the type.

src/platform/elements/form/Control.ts
```diff
@@ -234,372 +136,396 @@ export class NovoCustomControlContainerElement {
...
-        this.fieldInteractionApi.form = this.form;
-        this.fieldInteractionApi.currentKey = this.control.key;
-        try {
-          interaction.script(this.fieldInteractionApi, this.control.key);
-        } catch (err) {
-          console.info('Field Interaction Error!', this.control.key); // tslint:disable-line
-          console.error(err); // tslint:disable-line
+        this.templateContext.$implicit.tooltipPosition = this.tooltipPosition;
+        this.templateContext.$implicit.tooltip = this.tooltip;
+        this.templateContext.$implicit.startupFocus = this.control.startupFocus;//what does this do????
```
jgodi: This should focus the first field when the form is loaded, should see it on the form pages and should be autofocused

src/platform/elements/form/Control.ts
```diff
@@ -234,372 +136,396 @@ export class NovoCustomControlContainerElement {
...
+        this.templateContext.$implicit.fileBrowserImageUploadUrl = this.control.fileBrowserImageUploadUrl;//what does this do????
```
jgodi: Added for CKEditor, so images can be uploaded

src/platform/elements/form/Control.ts
```diff
@@ -234,372 +136,396 @@ export class NovoCustomControlContainerElement {
...
+        this.templateContext.$implicit.minimal = this.control.minimal;//what does this do????
```
jgodi: Used in the grouped forms for minimal styling (padding/label/etc)

src/platform/elements/form/Control.ts
```diff
@@ -234,372 +136,396 @@ export class NovoCustomControlContainerElement {
...
+        this.templateContext.$implicit.config = this.control.config;// IS THIS NEEDED???
```
jgodi: Not sure if it is

src/app/pages/elements/form/FormDemo.ts
amrutha-rajiv: sure

amrutha-rajiv: done

src/app/pages/elements/form/templates/VerticalDynamicForm.html
amrutha-rajiv: type has been removed now

PR #745, opened 2018-06-28
fix: datePickerInput events

ezrashim:
## **Description**
before:
- input event triggered both the calendar and form value update. This basically made it impossible for any user to key in values outside of a date picker.  

after:
- restrict input event to update the calendar value only and leave the form value. 
- add blur event to compute the form value when the blur event is initiated. 

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**
before:
employeedob

after:
screen shot 2018-06-28 at 11 38 26 am

### DIFF
--- a/src/platform/elements/date-picker/DatePickerInput.ts
+++ b/src/platform/elements/date-picker/DatePickerInput.ts
@@ -5,7 +5,7 @@ import { TAB, ENTER, ESCAPE } from '@angular/cdk/keycodes';
 // Vendor
 import { TextMaskModule } from 'angular2-text-mask';
 import * as dateFns from 'date-fns';
-import createAutoCorrectedDatePipe from 'text-mask-addons/dist/createAutoCorrectedDatePipe';	
+import createAutoCorrectedDatePipe from 'text-mask-addons/dist/createAutoCorrectedDatePipe';
 // App
 import { NovoDatePickerElement } from './DatePicker';
 import { NovoOverlayTemplateComponent } from '../overlay/Overlay';
@@ -23,7 +23,7 @@ const DATE_VALUE_ACCESSOR = {
   selector: 'novo-date-picker-input',
   providers: [DATE_VALUE_ACCESSOR],
   template: `
-        <input type="text" [name]="name" [(ngModel)]="formattedValue" [textMask]="maskOptions" [placeholder]="placeholder" (focus)="openPanel()" (keydown)="_handleKeydown($event)" (input)="_handleInput($event)" #input data-automation-id="date-input"/>
+        <input type="text" [name]="name" [(ngModel)]="formattedValue" [textMask]="maskOptions" [placeholder]="placeholder" (focus)="openPanel()" (keydown)="_handleKeydown($event)" (input)="_handleInput($event)" (blur)="_handleBlur($event)" #input data-automation-id="date-input"/>
         <i *ngIf="!hasValue" (click)="openPanel()" class="bhi-calendar"></i>
         <i *ngIf="hasValue" (click)="clearValue()" class="bhi-times"></i>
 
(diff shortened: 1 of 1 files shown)


src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -109,32 +107,65 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
     }
   }
 
+  _handleBlur(event: FocusEvent): void {
```
bvkimball: This looks the same as handleInput except for the the extra param. you should abstract this function.

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -109,32 +107,65 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
...
+    this._setFormValue(value);
+    this._changeDetectorRef.markForCheck();
+  }
+
+  private _setCalendarValue(value: any): void {
     if (value instanceof Date && this.value instanceof Date) {
       value = new Date(value.setHours(this.value.getHours(), this.value.getMinutes()));
     }
     this.value = value;
+    this._changeDetectorRef.markForCheck();
```
bvkimball: Since you call this in setTriggerValue, you should probably move this to be called in the dispatchOnChange.

src/platform/elements/date-picker/DatePickerInput.ts
ezrashim: resolved

src/platform/elements/date-picker/DatePickerInput.ts
ezrashim: resolved

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -95,46 +91,75 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
...
+  }
+
   registerOnChange(fn: (value: any) => {}): void {
     this._onChange = fn;
   }
   registerOnTouched(fn: () => {}) {
     this._onTouched = fn;
   }
-  public dispatchOnChange(newValue?: any, skip: boolean = false) {
+  public dispatchOnChange(newValue?: any, blur: boolean = false, skip: boolean = false) {
```
bvkimball: Seems like "skip" is never used anymore? should we remove it?

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -95,46 +91,75 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
...
+    this._handleEvent(event, true);
+  }
+
+  _handleEvent(event: Event, blur: boolean): void {
+    let value = (event.target as HTMLInputElement).value;
+    try {
+      let dateTimeValue = Date.parse(value);
+      if (!isNaN(dateTimeValue)) {
+        let dt = new Date(dateTimeValue);
+        if (dt !== this.value) {
```
bvkimball: This logic is duplicated is dispatchOnChange too.  I think we can clean the flow up more

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -95,46 +91,75 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
...
+          this.dispatchOnChange(dt, blur);
+        }
+      } else {
+        this.dispatchOnChange(null, blur);
+      }
+    } catch (err) {}
+    this.openPanel();
```
bvkimball: Are we sure blur events should open the panel?

src/platform/elements/date-picker/DatePickerInput.ts
```diff
@@ -95,46 +91,75 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
...
   }
 
+  _handleBlur(event: FocusEvent): void {
+    this._handleEvent(event, true);
+  }
+
+  _handleEvent(event: Event, blur: boolean): void {
+    let value = (event.target as HTMLInputElement).value;
+    try {
+      let dateTimeValue = Date.parse(value);
```
bvkimball: Maybe these few lines can go into a validateInput or parseValue function that checks parse, isNaN, and if dt !== this.value

PR #747, opened 2018-06-29
f/datePickerInput

ezrashim:
## **Description**
before:
- input event triggered both the calendar and the form value update. This made it impossible for any user to key in values outside of the date picker.  

after:
- restrict input event to update the calendar value only and leave the form value as is. 
- add blur event to compute the form value when the blur event is initiated. 

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**
before:
employeedob

after:
screen shot 2018-06-28 at 11 38 26 am


jgodi:
@ezrashim Do these same things need to happen for DateTimePickerInput?

ezrashim:
@jgodi changes to the datepicker should apply to dateTimePicker as well since dateTimePicker borrows the component from datepicker.

PR #751, opened 2018-07-11
WIP: feat(Value): add ability to retrieve rendered text from the parent component

tbcshifter:
## **Description**

Add a retrieveText trigger that calls the onRetrieveText handler that is passed the text rendered from the data.

#### **Verify that...**

- [ ] `npm run lint` passes

### DIFF
--- a/src/platform/elements/value/Value.ts
+++ b/src/platform/elements/value/Value.ts
@@ -1,9 +1,20 @@
 // NG2
-import { Component, Input, OnInit, HostBinding, OnChanges, SimpleChanges } from '@angular/core';
-//APP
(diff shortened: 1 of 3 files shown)


src/platform/elements/value/Value.ts
```diff
@@ -23,25 +34,33 @@ export enum NOVO_VALUE_THEME { DEFAULT, MOBILE };
...
+
   ngOnInit() {
     if (Helpers.isEmpty(this.meta)) {
       this.meta = {
-        label: ''
+        label: '',
       };
     }
+
+    this.retrieveText.subscribe((event) => {
```
jgodi: What does this do exactly? Passing in subjects can hurt performance, especially since you are subscribing and not unsubscribing. What is the actual goal of this?

PR #794, opened 2018-08-27
convert german values to am/pm

acrowe2:
## **Description**

Update time inputs to change vorm. to am and nachm. to pm.  The textMask on the time picker input was causing vorm. to appear as __ and nachm. to appear as am for the German locale.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/src/platform/elements/time-picker/TimePickerInput.ts
+++ b/src/platform/elements/time-picker/TimePickerInput.ts
@@ -201,6 +201,11 @@ export class NovoTimePickerInputElement implements OnInit, ControlValueAccessor 
       minute: '2-digit',
       hour12: !this.military,
     });
+    if (format.indexOf('vorm.') > -1) {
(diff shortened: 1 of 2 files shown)


src/platform/elements/time-picker/TimePickerInput.ts
```diff
@@ -201,6 +201,11 @@ export class NovoTimePickerInputElement implements OnInit, ControlValueAccessor
       minute: '2-digit',
       hour12: !this.military,
     });
+    if (format.indexOf('vorm.') > -1) {
```
jgodi: Shouldn't have locale specific code here. It should go into the i18n files and passed in via that.

bvkimball:
German Language by default should be using 24Hour Clock, researching if this is a real use case.

PR #797, opened 2018-08-29
feat(chips) lazy load chips data using getData

amrutha-rajiv:
## **Description**

- added ability to use getData from config to load chip data
- FI runs will be blocked until chip is in the stable state
#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


src/platform/elements/loading/Loading.ts
```diff
@@ -102,3 +102,24 @@ export class NovoSpinnerElement {
...
+            <path [ngStyle]="spinnerCss" class="spinner" d="M9.2,18.5C4.1,18.5,0,14.4,0,9.2S4.1,0,9.2,0c0.9,0,1.9,0.1,2.7,0.4c0.8,0.2,1.2,1.1,1,1.9
+                    c-0.2,0.8-1.1,1.2-1.9,1C10.5,3.1,9.9,3,9.2,3C5.8,3,3,5.8,3,9.2s2.8,6.2,6.2,6.2c2.8,0,5.3-1.9,6-4.7c0.2-0.8,1-1.3,1.8-1.1
+                    c0.8,0.2,1.3,1,1.1,1.8C17.1,15.7,13.4,18.5,9.2,18.5z"/>
+        </svg>
+    </div>
+    `,
+})
+export class NovoSimpleSpinnerElement {
+  @Input()
+  spinnerCss: any;
```
jgodi: Can this just be `fillColor`?

```
@Input() set fillColor(color) { // create css object }
```

src/platform/utils/form-utils/FormUtils.ts
```diff
@@ -440,7 +440,7 @@ export class FormUtils {
     return ret;
   }
 
-  toFieldSets(meta, currencyFormat, http, config: { token?: string; restUrl?: string; military?: boolean }, overrides?) {
+  toFieldSets(meta, currencyFormat, http, config: { token?: string, restUrl?: string, military?: boolean, formData?: any }, overrides?) {
```
jgodi: new property not used?

amrutha-rajiv: it's used in novo

src/platform/elements/loading/Loading.ts
amrutha-rajiv: sure

amrutha-rajiv: this is complete

src/platform/elements/loading/Loading.ts
```diff
@@ -102,3 +102,31 @@ export class NovoSpinnerElement {
     @Input() inverse: boolean;
     @Input() baseHref: string;
 }
+
+
+@Component({
```
bvkimball: why does the other spinner not work?

src/app/pages/elements/chips/ChipsDemo.ts
```diff
@@ -78,16 +87,18 @@ export class ChipsDemoComponent {
...
   private formatted: any;
   private async: any;
+  private asyncInitialize: any;
   private avalue: any;
   private placeholder: string = 'Select...';
   private value: any = ['Alabama'];
   private rowDemo: any;
   private rowValue: any;
-
+  asyncInitializeValue: any = [];
```
bvkimball: public or private

src/platform/elements/chips/Chips.ts
```diff
@@ -66,7 +80,7 @@ export class NovoChipElement {
 
 @Component({
   selector: 'chips,novo-chips',
-  providers: [CHIPS_VALUE_ACCESSOR],
+  providers: [CHIPS_VALUE_ACCESSOR, ChipsStateService],
```
bvkimball: Would it make sense to call this InputStateService? so that maybe it could be reusable?

amrutha-rajiv: yes. i was actually thinking similarly too.

src/platform/elements/loading/Loading.ts
amrutha-rajiv: the simple-spinner is smaller and more fitting for the showing the loading state within the picker. ALso, this was the UX preference

src/app/pages/elements/chips/ChipsDemo.ts
amrutha-rajiv: done

src/platform/elements/chips/Chips.ts
amrutha-rajiv: and done

jgodi:
Can we have a `ready` subject on each control that will fire when the control is ready for FI, then most of them will just fire immediately like right now and no FI needs to change. Then the FI that go against the async fields can subscribe to ready inside their FI and then handle it?

hiqbal01:
@jgodi I think the problem then would be if an interaction was executing against a field that is not a to-many picker, but inside the field interaction we try to fetch the value of a to-many field where the control is not ready, we would run into issues. Amrutha has enabled a form level ready state subject that bubbles up from input -> fieldset -> form so all the field interactions would only execute when all the controls are ready.

PR #818, opened 2018-09-21
WIP: F/validate integer input

monroepe:
## **Description**

Added a function to check the validity of integer input on blur. This is needed for validity checking on mobile.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


jgodi:
Why is this needed? If the control.type is number, then all you can type is numbers?

monroepe:
That is not the case on mobile for Android (at least with Ionic). It allows you to type non-number characters and after checking there is no current working way to block the input of characters because of issues with Ionic and Android keyboard. So currently it is an issue for mobile. At some point in the future we are planning to look into a different keyboard input, but since that is off a ways this is what we have to go with so that people don't enter bad input for integer fields.

PR #829, opened 2018-10-08
BH-57946: added Person as a linkable type

jay-quach-tbx:
Added Person as a linkable type.
Important, I'm not fully aware if making type Person linkable globally is a good idea or not. Need to answer that before merging


jay-quach-tbx:
Closed as this approach has issues

PR #835, opened 2018-10-22
Added person icon

jay-quach-tbx:
## **Description**

Allow to add an icon which represent the person type of novo-value:
image

### DIFF
--- a/projects/novo-examples/src/form-controls/value/formatter-value/formatter-value-example.ts
+++ b/projects/novo-examples/src/form-controls/value/formatter-value/formatter-value-example.ts
@@ -21,6 +21,8 @@ export class FormatterValueExample {
     associatedEntity: {
       entity: 'Placement',
     },
+    showEntityIcon: true,
+    entityIconClass: 'candidate',
     formatter: (value, args) => {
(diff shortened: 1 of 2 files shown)


projects/novo-examples/src/form-controls/value/formatter-value/formatter-value-example.ts
```diff
@@ -21,6 +21,8 @@ export class FormatterValueExample {
     associatedEntity: {
       entity: 'Placement',
     },
+    showEntityIcon: true,
```
jgodi: Do you mean to set the value or type here?

jay-quach-tbx: Are you referring to the line above or below? The `entityIconClass` was recommended from my team but I'm always open for further suggestions

jgodi: nvm I am dumb

jay-quach-tbx:
Found some blocking bugs in the base branch so we shouldn't merge. Still investigating

PR #838, opened 2018-10-22
Added person icon to value

jay-quach-tbx:
## **Description**

Allow to add an icon which represent the person type of novo-value:
image

### DIFF
--- a/src/platform/elements/value/Values.scss
+++ b/src/platform/elements/value/Values.scss
@@ -1,58 +1,62 @@
 novo-value {
+  display: flex;
+  flex-direction: row;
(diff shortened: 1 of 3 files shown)


src/platform/elements/value/Values.scss
```diff
@@ -1,58 +1,62 @@
...
-            i.#{$entity} {
-                color: $color;
-            }
-        }
-        i {
-            font-size: 1.1em;
-            margin-right: 6px;
-        }
+  }
+  .actions {
```
amrutha-rajiv: looks like this style got added twice?

PR #849, opened 2018-11-05
Fix(AddressControl & FieldInteractionAPI): Fix address disable state and promptUser data-automation-id

tb9will:
## **Description**

Fix FieldInteractionAPI promptUser function to pass data automation Id correctly. Update Address control to correctly update each address field's disabled state when the control is marked readOnly through field Interactions.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/form/extras/address/Address.ts
+++ b/projects/novo-elements/src/elements/form/extras/address/Address.ts
@@ -85,6 +85,17 @@ export interface NovoAddressConfig {
 export class NovoAddressElement implements ControlValueAccessor, OnInit {
   @Input()
   config: NovoAddressConfig;
+  private _readOnly: boolean = false;
+  @Input()
+  set readOnly(readOnly: boolean) {
+    this._readOnly = readOnly;
+    this.fieldList.forEach((field: string) => {
+      this.disabled[field] = this.readOnly;
+    });
+  }
+  get readOnly(): boolean {
+    return this._readOnly;
+  }
   states: Array<any> = [];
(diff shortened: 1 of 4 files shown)


projects/novo-elements/src/elements/form/extras/address/Address.ts
```diff
@@ -85,6 +85,17 @@ export interface NovoAddressConfig {
 export class NovoAddressElement implements ControlValueAccessor, OnInit {
   @Input()
   config: NovoAddressConfig;
+  private _readOnly: boolean = false;
+  @Input()
+  set readOnly(readOnly: boolean) {
+    this._readOnly = readOnly;
+    this.fieldList.forEach((field: string) => {
+      this.disabled[field] = this.readOnly;
+    });
```
amrutha-rajiv: this will interfere with some logic we have around disabling state if country isn't selected or if there are no states for a selected country. we need to account for that scenario as well.

bvkimball: how so?  If it is readOnly you can't change any values?

amrutha-rajiv: if you enable it then you can right, and the FI allows you to enable the address picker back

tb9will: You're right, that can leave the state picker enabled when it shouldn't be. Pushing up a fix now.

PR #850, opened 2018-11-06
stop execution if there are not any re-sizable columns

charlesabarnes:
## **Description**

This fixes the demo for data-tables

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [x] Run `BBO Automation`


projects/novo-elements/src/elements/data-table/data-table.component.ts
```diff
@@ -566,6 +566,9 @@ export class NovoDataTable<T> implements AfterContentInit, OnDestroy {
           );
         },
       );
+      if (!resizableColumns || resizableColumns.length === 0) {
```
bvkimball: instead of returning how about we only evaluate if opposite it true

PR #853, opened 2018-11-08
Update issue templates

akurtovic:
Add new Bug Report and Feature request templates


akurtovic:
Hard to tell from the pull request, but the end result should be that our Issues page looks something like this.

image

PR #860, opened 2018-11-14
Refactor BaseControl's types to be less redundant and more reuseable

DrewHoo:
## **Description**

This is a refactor for readability and usefulness of the `NovoControlConfig` type that reduces redundancy and makes the code cleaner.

1) `NovoControlConfig` and `BaseControl` have the same properties (aside from a few discrepancies which I will write about further down), except that `NovoControlConfig`'s properties are all optional. It is a common pattern to derive such classes so I made a base class called `Control`, which contains all the property information from both `NovoControlConfig` and `BaseControl`. I then derived `NovoControlConfig` from `Control` by using the `Partialify` mapped type. Now, the individual properties of `NovoControlConfig` and `BaseControl` are declared in only one place 🎸.
2) I alphabetized the types to make them more readable.
3) I discovered the following discrepancies between `NovoControlConfig` and `BaseControl`:  
 - `customControl` only existed in `NovoControlConfig`. I added it to `BaseControl`. It will still be accessible on `BaseControl`'s `__config` property.
- `BaseControl`'s type information for `layoutProperties` appears to be inaccurate; it is initialized to a property of the same name from the `NovoControlConfig` object that is passed to `BaseControl`'s constructor, but `NovoControlConfig`'s `layoutProperties` object has two additional properties: `customActions` and `edit`. I'm assuming this is an oversight and updated `BaseControl`'s type to reflect that.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**


projects/novo-elements/src/elements/form/controls/BaseControl.ts
```diff
@@ -16,158 +16,92 @@ export interface NovoGroupedControlConfig {
   initialValue?: {}[];
 }
 
-export interface NovoControlConfig {
-  validators?: Array<any>;
+class Control {
+  allowInvalidDate?: boolean;
+  appendToBody: boolean; // Deprecated;
```
amrutha-rajiv: i'm concerned that this className conflicts with the actual `NovoConrol` do we want to call this `ControlConfig`?

DrewHoo: That sounds great, I'll rename it `ControlConfig` 👍

PR #890, opened 2019-01-07
fix(Picker): Fixed potential memory leak issue in NovoPickerElement

ndickerson:
## **Description**

Our use of picker results overlay for pickers can in Windows 10 on some PCs in some situations result in the contents of the overlay being leaked. This problem is not present on a Mac, since the garbage collector within V8 will always reclaim almost all of the memory of the destroyed DOM elements. On Windows 10 specifically, for pickers with a lot of elements in the results, up to 90% of the DOM elements might be leaked, up to 50% of the time. 

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**
mac vs windows leak


ndickerson:
Thanks for the (tool) tip @charlesabarnes! Will make sure and test tooltips as well.

marissarcollins:
QA Complete, automation ran

ndickerson:
This change only affects pickers, and there was not an issue with overlays in general, so tooltips, dropdowns, etc. all have no changes.

PR #895, opened 2019-01-14
toast service to accept an optional custom toast element

acrowe2:
## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/toast/ToastService.ts
+++ b/projects/novo-elements/src/elements/toast/ToastService.ts
@@ -28,7 +28,7 @@ export class NovoToastService {
     this._parentViewContainer = view;
   }
 
-  alert(options) {
+  alert(options, toastElement: any = NovoToastElement) {
     return new Promise((resolve) => {
       if (!this._parentViewContainer) {
         console.error(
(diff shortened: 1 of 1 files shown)


projects/novo-elements/src/elements/toast/ToastService.ts
```diff
@@ -28,15 +28,15 @@ export class NovoToastService {
     this._parentViewContainer = view;
   }
 
-  alert(options) {
+  alert(options, toastElement: any = NovoToastElement) {
     return new Promise((resolve) => {
```
amrutha-rajiv: instead of `any` can we change it to a more specific type

DrewHoo: Ah, just saw this was merged. I was going to say: I can't figure out how to type this. The best I can come up with is `toastElement: Function`, because when you look at `NovoToastElement` (or any component) as a value, it looks like a function. That would at least prevent folks from passing plain objects or something else. It's not a perfect solution, though. It's very hard to type something based on a class's decorator (which I think would be the answer here, since the component decorator would tell us whether a class is a component) or based on an interface it implements (not ideal since I don't think you have to implement any interface in order to be a component).

PR #896, opened 2019-01-15
fix(datepicker): handle dates stored as strings properly

monroepe:
## **Description**

Handle dates that are stored as strings in (YYYY-MM-DD format) so that they do not get incorrectly converted.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/date-picker/DatePickerInput.ts
+++ b/projects/novo-elements/src/elements/date-picker/DatePickerInput.ts
@@ -56,32 +56,21 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor 
   /** View -> model callback called when autocomplete has been touched */
   _onTouched = () => {};
 
-  @Input()
-  name: string;
-  @Input()
-  start: Date;
-  @Input()
-  end: Date;
-  @Input()
-  placeholder: string;
-  @Input()
-  maskOptions: any;
-  @Input()
-  format: string;
-  @Input()
-  textMaskEnabled: boolean = true;
-  @Input()
-  allowInvalidDate: boolean = false;
+  @Input() name: string;
+  @Input() start: Date;
+  @Input() end: Date;
+  @Input() placeholder: string;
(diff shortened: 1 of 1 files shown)


projects/novo-elements/src/elements/date-picker/DatePickerInput.ts
```diff
@@ -234,6 +223,10 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
       if (!value) {
         return '';
       }
+      // Handle dates stored as string values
+      if (/\d{4}-\d{2}-\d{2}/.test(value)) {
+        return value;
+      }
```
amrutha-rajiv: unit tests for this would be great

projects/novo-elements/src/elements/date-picker/DatePickerInput.ts
```diff
@@ -234,6 +223,10 @@ export class NovoDatePickerInputElement implements OnInit, ControlValueAccessor
       if (!value) {
         return '';
       }
+      // Handle dates stored as string values
+      if (/\d{4}-\d{2}-\d{2}/.test(value)) {
```
akurtovic: Does `dateFns` not respect `YYYY-MM-DD` as a valid format? If it does, then this is essentially allowing this one format without having to pass a `userDefinedFormat`? And if we need this, what happens if it's `YYYY/MM/DD` or `YYYY\MM\DD`? And I'm pretty sure a value like `2012-99-00` is will come back as valid with this.

amrutha-rajiv: yes, and i just remembered that we have a version of the date-picker with text-mask enabled which allows only for `/` as a de-limiter. and the date-format is defined in the translation files. if we want to handle string values, why aren't we using the text mask?

PR #897, opened 2019-01-16
Fix(GroupedMultiPickerResults): update search/filter functionality

tb9-eli:
## **Description**

remove entity type from filtered values in search

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/picker/extras/grouped-multi-picker-results/GroupedMultiPickerResults.ts
+++ b/projects/novo-elements/src/elements/picker/extras/grouped-multi-picker-results/GroupedMultiPickerResults.ts
@@ -10,73 +10,90 @@ import { NovoLabelService } from '../../../../services/novo-label-service';
 @Component({
   selector: 'grouped-multi-picker-results',
   template: `
-        <div class="grouped-multi-picker-groups">
-            <novo-list direction="vertical">
-                <novo-list-item
-                    *ngIf="config.displayAll"
-                    (click)="selectCategory({ value: 'all', label: 'all' })"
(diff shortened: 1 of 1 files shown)


projects/novo-elements/src/elements/picker/extras/grouped-multi-picker-results/GroupedMultiPickerResults.ts
```diff
@@ -275,7 +292,14 @@ export class GroupedMultiPickerResults extends BasePickerResults implements OnIn
     if (this.searchTerm && this.searchTerm.length !== 0 && this.selectedCategory) {
       matches = matches.filter((match) => {
         const searchTerm = this.searchTerm.toLowerCase();
-        return match.label.toLowerCase().indexOf(searchTerm) > -1 || match.value.toLowerCase().indexOf(searchTerm) > -1;
+        return (
+          match.label.toLowerCase().indexOf(searchTerm) > -1 ||
+          match.value
+            .toLowerCase()
+            .split('.') // Split into to array since the first part of the string (of the format '[entityType].[value]') will break the filter functionality
+            .pop() // Only the last element of the array should be checked by the searchTerm filter, e.g. if the value is 'placement.FirstName' then 'placement' is not relevant to the users search
```
amrutha-rajiv: why are we always splitting the value by `.`?  if the match condition is different for the type of data you have with the picker, the match function should be overridden on the client side (in novo)

PR #903, opened 2019-01-28
feat(fieldInteractions): modified to make format optional

chasitywickenhauser-bh:
## **Description**

Adjusted field interactions api to allow config.format to be optional

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


projects/novo-elements/src/elements/form/FieldInteractionApi.ts
```diff
@@ -568,6 +567,9 @@ export class FieldInteractionApi {
             });
           },
         });
+        if (config.hasOwnProperty('format')) {
+          newConfig.format = config.format;
+        }
```
hiqbal01: Do we need to grab format from the existing control if one is not passed into modifyPickerConfig?

PR #929, opened 2019-03-23
ci(Cypress): Add cypress build step and begin automation on several demos

charlesabarnes:
## **Description**

I added a cypress build step and automate the dataTable and tiles features and start button and date picker demo automation,

Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**

image


charlesabarnes:
Screenshot added of a demo test.  Ready for review

cypress/support/commands.js
```diff
@@ -0,0 +1,27 @@
+// ***********************************************
```
akurtovic: looks like this whole file can go unless you're going to need it for something else soon.

cypress/integration/components/dataTable.js
```diff
@@ -0,0 +1,204 @@
...
+          .should('have.length', option);
+      })
+    });
+    it('should switch to the next page properly', () => {
+      cy.get(`${exampleContainerSelector} novo-tiles [data-automation-id="Basic"]`)
+        .eq(0)
+        .click()
+        .should('have.class', 'active');
+      cy.get(`${exampleContainerSelector} novo-data-table novo-data-table-pagination`)
+        .should('have.class', 'basic');
```
amrutha-rajiv: this is great! do we want to organize the code similar to how novo automation is done? so that we separate our the selectors into page classes, etc.

charlesabarnes: I _would_ agree, however due to how our demo is written, there is not much reuse of code outside of the file for the particular component/element.

amrutha-rajiv: it just makes it easier to read the code without worrying about which selector to change where as opposed to being able to reuse code.

PR #937, opened 2019-04-01
feat(AppBridge): App Bridge Fixes

ndickerson:
## **Description**

Fixing the AppBridge registration call to send back reject() when it fails, instead of the same response as when it succeeds. Also, adding type for preview slideout (Novo Only) and fixing the response of the DevAppBridge to not wrap the rest return in a second data object. 

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`


ndickerson:
Novo Automation Run Successfully on MR: https://bhsource.bullhorn.com/NOVO/novo/merge_requests/2220

PR #943, opened 2019-04-09
BH-61013 Novo: Using systemRequired in meta

tla-dev:
## **Description**

Meta will now return 3 field that indicate if a user is required to enter a value for a field

optional=false - hibernate required field, may be defaulted by backend code
required=true - user configurable required, Field Maps
systemRequired=true, Bullhorn defined required
Novo/Novo-elements needs to start honoring both systemRequired and *required *when handling required fields in forms.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


amrutha-rajiv:
looks good to me! we can merge it once it is tested.

PR #948, opened 2019-04-17
Update the default text on Picker field overlays

kvnfrederiksen:
## **Description**

I changed the default text on picker overlays so that users are not confused by the message, "No results to display..." as they focus on the field.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**


amrutha-rajiv:
lgtm

PR #949, opened 2019-04-17
fix(control-group) style & input type fixes

amrutha-rajiv:
## **Description**

- base-control: copying over subtype and type
- resetting edit/remove checks on every add and remove of controls
- fixed styles for required columns
- added the control key as a class to the label column
#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/form/ControlGroup.spec.ts
+++ b/projects/novo-elements/src/elements/form/ControlGroup.spec.ts
@@ -1,0 +1,117 @@
+// NG2
+import { ChangeDetectorRef } from '@angular/core';
+import { TestBed, async } from '@angular/core/testing';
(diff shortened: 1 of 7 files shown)


projects/novo-elements/src/elements/form/ControlGroup.spec.ts
```diff
@@ -0,0 +1,117 @@
+// NG2
+import { ChangeDetectorRef } from '@angular/core';
+import { TestBed, async } from '@angular/core/testing';
+import { FormBuilder, FormArray } from '@angular/forms';
+// App
```
DrewHoo: duplicate comment here, not sure if it's worth a fix

PR #955, opened 2019-04-29
added custom validation to file input control

chasitywickenhauser-bh:
## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**

fileInputCustomValidation


amrutha-rajiv:
nice!

projects/novo-elements/src/elements/form/extras/file/FileInput.ts
```diff
@@ -239,16 +306,30 @@ export class NovoFileInputElement implements ControlValueAccessor, OnInit, OnDes
     this.process(Array.from(event.target.files));
   }
 
+  validate(files): boolean {
+    let passedValidation = true;
+    if (this.layoutOptions.customValidation) {
+      this.layoutOptions.customValidation
+        .filter((validation) => validation.action === 'upload')
+        .forEach((uploadValidation) => {
```
hiqbal01: Should we be using a find instead of a for each to stop checking validations as soon as any of them fails?

hiqbal01: or do we want them all to run regardless?

chasitywickenhauser-bh: Good Question. I went with a forEach so that it would loop through everything in case the validation methods were outputting information so that the user would know everything that was wrong. I can easily be persuaded to exit at the first failure though.

DrewHoo: whichever way we decide this, it's worth noting that the check gets short-circuited as soon as one is false (thus `passedValidation` becomes false, so you wouldn't be able to see failures after the first.

chasitywickenhauser-bh: I updated with making the filter and foreach separate so that it will not short-circuit.

PR #957, opened 2019-05-01
BH-61405 SPECIALIZED_OPTIONS is a select

tla-dev:
## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/utils/form-utils/FormUtils.ts
+++ b/projects/novo-elements/src/utils/form-utils/FormUtils.ts
@@ -128,6 +128,7 @@ export class FormUtils {
       'HTML-MINIMAL': 'editor-minimal',
       YEAR: 'year',
       WORKFLOW_OPTIONS: 'select',
+      SPECIALIZED_OPTIONS: 'select',
     };
     let dataTypeToTypeMap = {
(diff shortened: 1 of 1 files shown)


projects/novo-elements/src/utils/form-utils/FormUtils.ts
```diff
@@ -587,6 +588,8 @@ export class FormUtils {
       return [{ value: false, label: this.labels.no }, { value: true, label: this.labels.yes }];
     } else if (field.workflowOptions && fieldData) {
       return this.getWorkflowOptions(field.workflowOptions, fieldData);
+    } else if (field.dataSpecialization === 'SPECIALIZED_OPTIONS') {
+      return field.options;
```
amrutha-rajiv: we should be filtering out options with isEditOption = false

tla-dev: is isEditOption going to be implemented in the future? Here is a current example {value: 3, label: "Weeks", readOnly: false}

PR #959, opened 2019-05-03
Add filter box for filtering multi-select columns

trenthibbard:
## **Description**

In data-table on multi-select columns, there is now a text box to search for options within the options list.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**
image


projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
```diff
@@ -283,6 +304,19 @@ export class NovoDataTableCellHeader<T> implements IDataTableSortFilter, OnInit,
     this.multiSelect = this.config.filterConfig && this.config.filterConfig.type ? this.config.filterConfig.type === 'multi-select' : false;
     if (this.multiSelect) {
       this.multiSelectedOptions = this.filter ? [...this.filter] : [];
+      if (typeof this.config.filterConfig.options[1] === 'string') {
+        (this.config.filterConfig.options as string[]).forEach(
```
hiqbal01: Could you use the map function here, i.e.
`this.multiSelectedOptionIsHidden = (this.config.filterConfig.options as string[]).map((option: string) => { option, hidden: false });`

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
```diff
@@ -283,6 +304,19 @@ export class NovoDataTableCellHeader<T> implements IDataTableSortFilter, OnInit,
...
+          (option: string): void => {
+            this.multiSelectedOptionIsHidden.push({ option: option, hidden: false });
+          },
+        );
+      } else {
+        (this.config.filterConfig.options as IDataTableColumnFilterOption[]).forEach(
```
hiqbal01: same as above

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
DrewHoo: `options` has a well defined type, but you're casting it here in order to narrow the type. TypeScript should see the typeof check and type narrow for you, making the cast unnecessary. In this case though you have to use bracket notation and that's breaking type checking. To avoid casting, move your typeof check inside of the forEach loop:
```
this.config.filterConfig.options.forEach((s: string | IDataTableColumnFilterOption) => {
        if (typeof s === 'string') {
             //string stuff
        } else {
             // IDataTableColumnFilterOption stuff
        }
      });
```
note that filterConfig & options are optional and therefore need undefined checks before accessing them (we do not yet have the strict TS compiler options enabled that should make accesses like this illegal).

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
```diff
@@ -326,12 +368,79 @@ export class NovoDataTableCellHeader<T> implements IDataTableSortFilter, OnInit,
...
     this.multiSelectedOptions = this.filter ? [...this.filter] : [];
     this.dropdown.closePanel();
+    this.clearOptionFilter();
   }
 
   public filterMultiSelect(): void {
-    let actualFilter = this.multiSelectedOptions.length > 0 ? [...this.multiSelectedOptions] : undefined;
-    this.filterData(actualFilter);
-    this.dropdown.closePanel();
+    if ((this.multiSelectedOptions.length === 0 && !this.filter) || this.multiSelectedOptions === this.filter) {
```
DrewHoo: If `this.multiSelectedOptions` is an array, the equality comparison against `this.filter` will always fail, and you need to use a helper function to sort & compare the elements of each array. Gonna DM you some novo code that handles this, if this turns out to be a necessary check.

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
trenthibbard: That actually wasn't working for me, I get ```error TS2349: Cannot invoke an expression whose type lacks a call signature. Type '((callbackfn: (value: string, index: number, array: string[]) => void, thisArg?: any) => void) | ((callbackfn: (value: IDataTableColumnFilterOption, index: number, array: IDataTableColumnFilterOption[]) => void, thisArg?: any) => void)' has no compatible call signatures.```

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
trenthibbard: Thanks for calling this out, it actually isn't necessary now that I think about it

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
trenthibbard: Certainly

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
```diff
@@ -326,12 +368,79 @@ export class NovoDataTableCellHeader<T> implements IDataTableSortFilter, OnInit,
...
+  public multiSelectOptionIsHidden(option: string | IDataTableColumnFilterOption): boolean {
+    return this.multiSelectedOptionIsHidden.find((record) => record.option === option).hidden;
+  }
+
+  public multiSelectHasVisibleOptions(): boolean {
+    return this.multiSelectedOptionIsHidden.some((record) => !record.hidden);
+  }
+
+  private getOptionText(option: string | IDataTableColumnFilterOption): string {
+    if (typeof option !== 'object') {
```
DrewHoo: I think it'd be preferable to type narrow based on a property that would be in IDataTableColumnFilterOption instead.  like `'label' in object` since label is the only required property on that interface.

projects/novo-elements/src/elements/form/extras/file/FileInput.ts
```diff
@@ -239,16 +306,30 @@ export class NovoFileInputElement implements ControlValueAccessor, OnInit, OnDes
     this.process(Array.from(event.target.files));
   }
 
+  validate(files): boolean {
+    let passedValidation = true;
+    if (this.layoutOptions.customValidation) {
+      this.layoutOptions.customValidation
```
DrewHoo: You could make this method a lot shorter by using `.every` here instead of `.forEach` since it would let you return a chain of array calls

trenthibbard: This is in Master already, different ticket

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
trenthibbard: Can't do an `in` check on object because it could be a string

PR #968, opened 2019-05-16
BH-61236 field hidden when error message

tla-dev:
## **Description**

Fields can be hidden in a form because an other fields. Those hidden fields should stay hidden if the user go back and forth in the error page.

#### **Verify that...**

- [*] Any related demos were added and `npm start` and `npm run build` still works
- [*] `npm run lint` passes
- [*] `npm test` passes and code coverage is increased
- [*] `npm run build` still works

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/form/NovoFormControl.ts
+++ b/projects/novo-elements/src/elements/form/NovoFormControl.ts
@@ -1,9 +1,8 @@
 // NG2
-import { FormControl, FormGroup, Validators } from '@angular/forms';
+import { FormControl, Validators } from '@angular/forms';
 import { EventEmitter } from '@angular/core';
 // APP
 import { NovoControlConfig } from './FormControls';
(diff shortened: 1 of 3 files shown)


projects/novo-elements/src/elements/form/NovoFormControl.ts
```diff
@@ -62,6 +61,7 @@ export class NovoFormControl extends FormControl {
   checkboxLabel?: string;
   restrictFieldInteractions?: boolean;
   warning?: string;
+  isHiddenByLogic = false; // if a control is hidden because not releavant (email address in billing profile)
   private historyTimeout: any;
```
amrutha-rajiv: This might not be the way to hide the controls

tla-dev: yep, I changed the code accordingly.

PR #971, opened 2019-05-21
F/bh 61437 wfr pickers

DrewHoo:
## **Description**
The goal of these changes it to make it possible for WFR pickers to search on an empty search term when the user clicks on a WFR entity picker.

There are three significant changes to make this possible:
1) On picker configs, a `minSearchLength` of `0` is now possible (see BasePickerResults `search` function).
2) In FormUtils `getControlForField`, field configs are no longer completely replaced by the optionsConfig returned from 'getControlOption`. Properties with the same name will still be overwritten. This is designed to allow Novo to add some picker config in field meta before using FormUtils to create picker controls.
3) In FieldInteractionAPI, modifyPickerConfig now preserves the `minSearchLength` property on picker configs. I wonder if it should just be preserving the picker config in general but doing so would add too much testing scope to this feature.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**


projects/novo-elements/src/elements/picker/extras/base-picker-results/BasePickerResults.ts
```diff
@@ -66,7 +66,7 @@ export class BasePickerResults {
   }
 
   set term(value) {
-    if (value !== this._term || this.page === 0) {
+    if (value !== this._term || this.page === 0 || value === '') {
```
DrewHoo: @amrutha-rajiv this is the fix I found for the issue I showed you. I'm not sure if there is a better approach. The `billingProfile` picker behaves nicely because it retains its `this._term` and so it enters this if block when `value` is an empty string. 
Whereas for location, `this._term` is an empty string *before the user searches a second time*, so it does not enter this if block to process the search. My hunch is that this is a result of the field interaction that clears out `billingLocation` and is somehow resetting the `this._term` to an empty string. 
I'm not sure what else I could compare on here because the property that changes (and should indicate that we need to do a new search) is `[picker].config.options`, and it is simply replaced with a new function that has a different filter.

DrewHoo: An alternative to this change might be to override `term`'s setter in `location-picker-results` to do some kind of a diff on the options function? LMK what you think.

projects/novo-elements/src/elements/form/FieldInteractionApi.ts
```diff
@@ -537,33 +537,33 @@ export class FieldInteractionApi {
...
-      let newConfig: any = {
+
+      const newConfig: NovoControlConfig['config'] = {
+        ...(enableInfiniteScroll && { enableInfiniteScroll }),
+        ...(Number.isInteger(minSearchLength) && { minSearchLength }),
         resultsTemplate: control.config.resultsTemplate,
       };
       if (config.optionsUrl || config.optionsUrlBuilder || config.optionsPromise) {
-        newConfig = Object.assign(newConfig, {
+        Object.assign(newConfig, {
```
hiqbal01: Should we be using the spread operator here?

projects/novo-elements/src/elements/picker/extras/base-picker-results/BasePickerResults.ts
```diff
@@ -66,7 +66,7 @@ export class BasePickerResults {
   }
 
   set term(value) {
-    if (value !== this._term || this.page === 0) {
+    if (value !== this._term || this.page === 0 || (this.page > 0 && value === '')) {
```
DrewHoo: refactor into a named method for clarity

projects/novo-elements/src/elements/picker/extras/base-picker-results/BasePickerResults.ts
```diff
@@ -134,14 +134,16 @@ export class BasePickerResults {
             this.isStatic = true;
             // Arrays are returned immediately
             resolve(this.structureArray(options));
-          } else if (term && term.length >= (this.config.minSearchLength || 1)) {
+          } else if ((this.config.minSearchLength === 0) || (term && term.length >= (this.config.minSearchLength || 1))) {
```
DrewHoo: remove the `|| 1` in this expression

PR #973, opened 2019-05-28
feat(data-table): added cell renderer to support decimal and negative…

chasitywickenhauser-bh:
… decimal value styling

## **Description**

I added a formatted column for "big decimal". Shows two decimal points and negative values are shown within parens instead of having the leading minus signs.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**
bigDecimalFormattedColumn


projects/novo-elements/src/services/novo-label-service.ts
```diff
@@ -202,6 +202,20 @@ export class NovoLabelService {
     return new Intl.NumberFormat(this.userLocale, options).format(value);
   }
 
+  formatBigDecimal(value: number): string {
+    let valueAsString = value.toString();
```
hiqbal01: do we need to worry about nulls

chasitywickenhauser-bh: I'll add a check

PR #980, opened 2019-06-07
feat(pickers): enable FieldInteractionAPI to create filtered options calls for pickers, allow pickers to infinitely scroll with zero-length search term

DrewHoo:
## **Description**

FieldInteractionAPI's `modifyPickerConfig` can now create picker `options` functions (which are called in `BasePickerResults.search`) by calling an optional `filteredOptionsCreator` function that exists on picker configs. This allows pickers' own configs to determine how the picker is allowed to be changed, rather than having field interactions violate encapsulation.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -25,7 +25,7 @@
       true,
       "spaces"
     ],
-    "interface-over-type-literal": true,
+    "interface-over-type-literal": false,
     "label-position": true,
     "member-access": false,
     // "member-ordering": [
--- a/tslint.json
+++ b/tslint.json
@@ -88,7 +88,8 @@
     // "radix": true,
     "semicolon": [
       true,
-      "always"
+      "always",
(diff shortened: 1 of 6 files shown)


tslint.json
```diff
@@ -88,7 +88,8 @@
     // "radix": true,
     "semicolon": [
       true,
-      "always"
+      "always",
+      "ignore-bound-class-methods"
```
DrewHoo: If you have an arrow function bound to a class, prettier thinks you should have a semicolon on the end of it, and tslint does not. This causes chaos in any editor that uses both. This change will cause tslint to allow semicolons to follow arrow functions bound to classes.

PR #986, opened 2019-06-13
Merge provided config with base config in CKEditor

trenthibbard:
## **Description**

In https://github.com/bullhorn/novo-elements/pull/982, we enabled the ability to pass a config object to Editor Control to update the underlying CKEditor options. However, in its current implementation, providing a config object will completely replace the base config defined in novo-elements. To simplify changing a small number of attributes, I think it would be better if we still loaded the base config and treated the provided config as an override for the fields provided..

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**
image

### DIFF
--- a/projects/novo-elements/src/elements/ckeditor/CKEditor.ts
+++ b/projects/novo-elements/src/elements/ckeditor/CKEditor.ts
@@ -83,7 +83,7 @@ export class NovoCKEditorElement implements OnDestroy, AfterViewInit, ControlVal
   }
 
   ngAfterViewInit() {
-    let config = this.config || this.getBaseConfig();
+    let config = Object.assign(this.getBaseConfig(), this.config);
     if (this.startupFocus) {
       config.startupFocus = true;
     }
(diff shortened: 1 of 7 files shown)


DrewHoo:
This change makes a lot of sense to me, but wanted to mention this is potentially a breaking change bc the base config made in CKEditor is disabling a lot of stuff. I'm interested to know what other folks think.

trenthibbard:
> This change makes a lot of sense to me, but wanted to mention this is potentially a breaking change bc the base config made in CKEditor is disabling a lot of stuff, and could disable things that people passing configs into novo-editor don't want disabled. I'm interested to know what other folks think.
> 
> Edit: @trenthibbard pointed out this would only be breaking for someone using novo-editor by itself; dynamic forms were unable to pass configs to novo-editor until yesterday.

Details on the CKEditor config object are here: https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html

The fields we use in our base novo-editor config that differ from the default CKEditor config are:
enterMode (default:CKEDITOR.ENTER_P, novo-editor: CKEDITOR.ENTER_BR)
shiftEnterMode (default: CKEDITOR.ENTER_BR, novo-editor: CKEDITOR.ENTER_P)
disableNativeSpellChecker (default: true, novo-editor: false)
removePlugins (default: '', novo-editor: 'liststyle,tabletools,contextmenu')
extraAllowedContent (default: null, novo-editor: '*(*){*};table tbody tr td th[*];')
font_names (novo-editor is the default list plus Calibri)
toolbar (default: has a lot more buttons, novo-editor: depends on "minimal" directive)
filebrowserImageUploadUrl (default: '' (disabled), novo-editor: this.fileBrowserImageUploadUrl (unless [minimal] is true))

trenthibbard:
Since this will be a breaking change, we will not do this.

PR #987, opened 2019-06-14
changing opacity on picker lists

awuBullhorn:
## **Description**

Small css change to up opacity on picker items


marissarcollins:
qa passing

PR #989, opened 2019-06-18
BH-61571: Hide char count for field if not focused

kvnfrederiksen:
## **Description**

Character counts for fields that use them should only ever display if the field is focused.

This small change in logic ensures that the focus state is checked before anything else.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/form/Control.ts
+++ b/projects/novo-elements/src/elements/form/Control.ts
@@ -267,7 +267,7 @@ export class NovoControlElement extends OutsideClick implements OnInit, OnDestro
     const MAX_LENGTH_CONTROL_TYPES: string[] = ['textbox', 'picker', 'text-area'];
     let charCount: boolean =
       this.focused &&
-      this.form.controls[this.control.key].maxlength &&
+      !!this.form.controls[this.control.key].maxlength &&
       MAX_LENGTH_CONTROL_TYPES.includes(this.form.controls[this.control.key].controlType);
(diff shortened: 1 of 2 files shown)


projects/novo-elements/src/elements/form/Control.ts
```diff
@@ -265,11 +265,13 @@ export class NovoControlElement extends OutsideClick implements OnInit, OnDestro
...
     let charCount: boolean =
-      (this.form.controls[this.control.key].maxlength &&
-        this.focused &&
-        (this.form.controls[this.control.key].controlType === 'text-area' ||
-          this.form.controls[this.control.key].controlType === 'textbox')) ||
-      (this.form.controls[this.control.key].maxlength && this.form.controls[this.control.key].controlType === 'picker');
+      this.focused &&
+      this.form.controls[this.control.key].maxlength &&
+      (
+        this.form.controls[this.control.key].controlType === 'text-area' ||
```
DrewHoo: Loving this boolean simplification. I think it might be clearer if it were an `Array.prototype.includes` check for those values instead of accessing controlType for each check i.e. `['textbox', 'picker', 'text-area'].includes(controlType)`

kvnfrederiksen: Got that array in a const for readability now.

PR #993, opened 2019-06-25
onChange for ControlGroup

tla-dev:
## **Description**

Propagate the onChange() of a novo-control to the novo-control-group

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/form/ControlGroup.ts
+++ b/projects/novo-elements/src/elements/form/ControlGroup.ts
@@ -48,8 +48,7 @@ export class NovoControlGroup implements AfterContentInit, OnChanges {
   }
   private _vertical: boolean = false;
   // Hides/shows the add button for adding a new control
-  @Input()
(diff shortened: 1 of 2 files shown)


projects/novo-elements/src/elements/form/ControlGroup.ts
```diff
@@ -170,6 +157,10 @@ export class NovoControlGroup implements AfterContentInit, OnChanges {
     }
   }
 
+  onChange(change: any): void {
+    this.change.emit(this);
+  }
```
amrutha-rajiv: do we need to emit `this`? why not just emit the change?

tla-dev: the change is the element itself (the picker, dropdown, date...). It doesn't have the index

PR #1000, opened 2019-07-09
Accept Custom Preselected Config

MichaelDill:
## **Description**

Allow Picker results to accept a custom preselected function for determining whether a result has already been selected.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/picker/extras/base-picker-results/BasePickerResults.ts
+++ b/projects/novo-elements/src/elements/picker/extras/base-picker-results/BasePickerResults.ts
@@ -382,6 +382,14 @@ export class BasePickerResults {
   }
 
   preselected(match) {
+    if (this.config.preselected) {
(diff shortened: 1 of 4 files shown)


projects/novo-elements/src/elements/picker/extras/base-picker-results/BasePickerResults.ts
```diff
@@ -382,6 +382,14 @@ export class BasePickerResults {
   }
 
   preselected(match) {
+    if (this.config.preselected) {
+      let preselectedFunc: Function = this.config.preselected;
+      return (
+        this.selected.findIndex((item) => {
```
DrewHoo: If you don't need the index other than to tell whether a value has been found in the array, it'd be more idiomatic to use `Array.prototype.some` with that callback

PR #1003, opened 2019-07-11
feat(picker) non-destructively add properties to picker configs via field interactions

DrewHoo:
## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/form/FieldInteractionApi.ts
+++ b/projects/novo-elements/src/elements/form/FieldInteractionApi.ts
@@ -570,7 +570,7 @@ export class FieldInteractionApi {
     }
 
     const config = {
-      ...(control && control.config),
+      ...control.config,
       ...properties,
     };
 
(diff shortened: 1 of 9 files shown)


projects/novo-elements/src/elements/form/FieldInteractionApi.ts
```diff
@@ -561,6 +562,21 @@ export class FieldInteractionApi {
...
   }
+
+  addPropertiesToPickerConfig(key: string, properties: { [key: string]: unknown }) {
+    let control = this.getControl(key);
+    if (!control || control.restrictFieldInteractions) {
+      return;
+    }
+
+    const config = {
+      ...(control && control.config),
```
ndickerson: The use of `control &&` seems unnecessary, since you have the `if (!control ||` above.

DrewHoo: good catch!

PR #1016, opened 2019-07-29
Change how dirty flag is being calculated

awuBullhorn:
## **Description**

Fixed bug where some fields don't calculate dirty flag properly, specifically booleans set to false, and integers set to 0

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/form/controls/BaseControl.ts
+++ b/projects/novo-elements/src/elements/form/controls/BaseControl.ts
@@ -121,7 +121,7 @@ export class BaseControl extends ControlConfig {
     this.metaType = config.metaType;
     this.placeholder = config.placeholder || '';
     this.config = config.config || null;
-    this.dirty = !!config.value;
+    this.dirty = !!(config.value || (config.dataType === 'Integer' && config.value !== null));
     this.multiple = !!config.multiple;
     this.headerConfig = config.headerConfig || null;
     this.currencyFormat = config.currencyFormat || null;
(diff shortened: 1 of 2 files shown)


marissarcollins:
BBO automation is passing

projects/novo-elements/src/elements/form/controls/BaseControl.ts
```diff
@@ -121,7 +121,7 @@ export class BaseControl extends ControlConfig {
     this.metaType = config.metaType;
     this.placeholder = config.placeholder || '';
     this.config = config.config || null;
-    this.dirty = !!config.value;
+    this.dirty = !!config.value || (config.dataType === 'Integer' && config.value !== null);
     this.multiple = !!config.multiple;
```
amrutha-rajiv: what is the need to add a dataType specific check here?

awuBullhorn: When the field is an integer and the default value is 0, it's not being marked dirty because the dirty flag evaluates to false. Then when the form is saved, that default int value doesn't get saved.

amrutha-rajiv: i think we need to change the check for the `this.dirty` to make sure it doesn't incorrectly assume that `0` is an empty value

amrutha-rajiv:
looks good to me!

PR #1028, opened 2019-08-22
feat(html) add pendo tags

JonathanBraun:
## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell.component.ts
```diff
@@ -193,6 +195,9 @@ export class NovoDataTableCellHeader<T> implements IDataTableSortFilter, OnInit,
   @Input()
   allowMultipleFilters: boolean = false;
 
+  @Input()
+  dataFeatureId: string;
+
```
amrutha-rajiv: wondering if you can directly use `column.id` instead of passing it through this input.. is that value available to the template?

JonathanBraun: cool, i'll give that a shot

PR #1048, opened 2019-10-07
BH-63735 adding pagination related to modify picker config

ShylendraPandravisam:
## **Description**
when a picker config is overridden by modify picker config. code was not respecting pagination, upon scrolling down was always fetching same exact first (20) results and appending to exiting. so adding page option to use further when forming an url.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


ShylendraPandravisam:
NO LOnger needed this change

PR #1060, opened 2019-12-13
Style tabbed group picker for low resolution

kvnfrederiksen:
## **Description**

The tabbed-group-picker currently changes position if the user is at minimum supported resolution (1280x720). This simple change to the CSS file allows for the element to stay on the page and work as intended at this resolution.

#### **Verify that...**

- [x ] Any related demos were added and `npm start` and `npm run build` still works
- [x ] `npm run lint` passes
- [x ] `npm run build` still works

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**

You can see here, at 1280x720 resolution, that the drop-down moves up over the field, so the user cannot dynamically see their changes in the field itself.

image


projects/novo-elements/src/elements/tabbed-group-picker/TabbedGroupPicker.scss
```diff
@@ -145,11 +154,14 @@
...
+              margin-right: 0;
+              label span {
+                white-space: nowrap;
+                overflow: hidden;
+                text-overflow: ellipsis;
+                width: 16.23em;
+              }
             }
           }
         }
```
kvnfrederiksen: Check for empty state as well

projects/novo-elements/src/elements/tabbed-group-picker/TabbedGroupPicker.scss
```diff
@@ -127,11 +130,17 @@
...
-        height: 100%;
+        @media (min-height: 720px) {
+          height: 100%;
+        }
 
         cdk-virtual-scroll-viewport {
           height: 27em;
           overflow-x: hidden;
+          @media (max-height: 720px) {
+            height: 19em;
```
kvnfrederiksen: Try setting this height to a vh

PR #1061, opened 2019-12-19
WIP: Tiles with Options and Inline-Embedded controls

chasitywickenhauser-bh:
## **Description**

Added support for Tiles with options (other than true/false).

Added support for inline-embedded controls

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
- [x] Run `BBO Automation`

##### **Screenshots**
image


projects/novo-elements/src/utils/form-utils/FormUtils.ts
```diff
@@ -795,4 +810,21 @@ export class FormUtils {
...
+          let [parentFieldName, fieldName] = field.split('.');
+          if (!data[parentFieldName]) {
+              data[parentFieldName] = {};
+          }
+          data[parentFieldName][fieldName] = data[field];
+          delete data[field];
+      });
+    }
+
+    return data;
```
amrutha-rajiv: is this function being called by anything?

amrutha-rajiv: can you please remove `examples.module.ts` from your commit? this is auto generated. we'll probably want to add this to our .gitignore as well.

chasitywickenhauser-bh: Yes, the method is used for embedded and inline_embedded fields. I have removed examples.module.ts and added it to the gitignore file.

PR #1063, opened 2019-12-31
BH-64754: Add for new novo version

kvnfrederiksen:
## **Description**

This will make sure the hotfix applied in 3.16.1 will also publish to the new version of novo in the 3.17.0 release.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [x] Run `BBO Automation`

### DIFF
--- a/projects/novo-elements/src/elements/data-table/data-table.pipes.ts
+++ b/projects/novo-elements/src/elements/data-table/data-table.pipes.ts
@@ -46,7 +46,8 @@ export class DateTableDateTimeRendererPipe<T> implements PipeTransform {
   constructor(private labels: NovoLabelService) {}
   transform(value: any, column: IDataTableColumn<T>): string {
     if (!Helpers.isEmpty(value)) {
-      return column.format ? value : this.labels.formatDate(interpolateCell<T>(value, column));
+      let val = interpolateCell<T>(value, column);
+      return this.labels.formatDateShort(val);
     }
     return '';
   }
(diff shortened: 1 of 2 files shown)


projects/novo-elements/src/elements/data-table/data-table.pipes.ts
```diff
@@ -46,7 +46,8 @@ export class DateTableDateTimeRendererPipe<T> implements PipeTransform {
   constructor(private labels: NovoLabelService) {}
   transform(value: any, column: IDataTableColumn<T>): string {
     if (!Helpers.isEmpty(value)) {
-      return column.format ? value : this.labels.formatDate(interpolateCell<T>(value, column));
+      let val = interpolateCell<T>(value, column);
+      return this.labels.formatDateShort(val);
     }
```
amrutha-rajiv: please add unit tests

kvnfrederiksen: Done!

projects/novo-elements/src/elements/data-table/data-table.pipes.ts
```diff
@@ -46,7 +46,8 @@ export class DateTableDateTimeRendererPipe<T> implements PipeTransform {
   constructor(private labels: NovoLabelService) {}
   transform(value: any, column: IDataTableColumn<T>): string {
     if (!Helpers.isEmpty(value)) {
-      return column.format ? value : this.labels.formatDate(interpolateCell<T>(value, column));
+      let val = interpolateCell<T>(value, column);
```
tbo47: Can you use const instead of let here?

kvnfrederiksen: That's a possibility. However, this is for a hotfix, and is a simple reversion to what was there before. As such, I'd rather play it safe with what I know works and move forward quickly. A follow-on ticket will be set up to address concerns such as these.

PR #1065, opened 2020-01-10
update date-fns

tbo47:
## **Description**

update date-fns 1 -> 2

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


bvkimball:
@tbo47  Is this still planned? if so we need to update the following as well:

- [ ] `change import * as dateFns from 'date-fns';` to `import { format } from 'date-fns';`
- [ ] interval functions were rename `eachDay` is now `eachDayOfInterval`
- [ ] range and interval function now use a new interface like { start: new Date(2014, 0, 10), end: new Date(2014, 0, 20) }
- [ ] All functions now only accept Date or Number not strings, we would need to know if it is a string and parse it first with `parseIso`

The last task here is the hardest because in general we expected the dateFns library to do this check, our code intentionally ignores needing to know the state of the date as we translate from the input or the picker.

PR #1109, opened 2020-05-21
BH-66970 refresh dropdown list array when it changes

dvoegelin:
## **Description**

subscribing to changes on QueryList so that we can reset/reload the `items` property dynamically, since we now have a component that can actively filter dropdown results.

also resetting `activeIndex` when we do this, since we use that to access `items` array elements and will throw errors if we get out of range with shorter, filtered items lists.

QueryList also unsubscribes itself on destroy:
https://github.com/angular/angular/blob/7d137d7f8872a6fba72668e32f9baf2c5dcfc48b/packages/core/src/linker/query_list.ts#L115

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/dropdown/Dropdown.ts
+++ b/projects/novo-elements/src/elements/dropdown/Dropdown.ts
@@ -25,7 +25,12 @@ import { notify } from '../../utils/notifier/notifier.util';
   template: `
     <ng-content select="button" #trigger></ng-content>
     <novo-overlay-template [parent]="element" [width]="width" [position]="side" [scrollStrategy]="scrollStrategy">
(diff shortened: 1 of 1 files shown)


projects/novo-elements/src/elements/dropdown/Dropdown.ts
```diff
@@ -288,6 +294,9 @@ export class NovoListElement implements AfterContentInit {
 
   public ngAfterContentInit(): void {
     this.dropdown.items = this.items;
+    this.items.changes.subscribe(() => {
+      this.dropdown.items = this.items;
+    });
   }
```
amrutha-rajiv, line 300: please add an unsubscribe as well.

PR #1115, opened 2020-05-27
BH-62199 prefixing truncated bigdecimal values with a ~ to convey tha…

dvoegelin:
…t they are approx

## **Description**

if we are truncating values to 2 decimal places, we want a way to convey to the user that the displayed values are approximations of the stored values, so we will display a '~' in front of the values in this case.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


dvoegelin:
decided to not add tildes

Issue #1121, opened 2020-06-19
Chips are stacking vertically instead of horizontal


blitzmann:
**Describe the bug**
It looks like the chips component is stacking vertically instead of horizontally

**To Reproduce**
1) Use chips component and put multiple values in
2) ???
3) Profit!

**Expected behavior**
Chip values should be horizontal

**Actual behavior**
Chip values are stacked vertically

**Specifications**
  - Version: 427fa8e (latest, but also seems to be a thing at least as far back as v4.1.0 tag)

**Screenshots**
Demo site shows chips stacking vertically:
![image](https://user-images.githubusercontent.com/3904767/85173570-86a80c80-b241-11ea-8477-99ae9d1f3193.png)

Bullhorn production shows that chips are horizontal:
![image](https://user-images.githubusercontent.com/3904767/85173378-231ddf00-b241-11ea-8a90-25fad85ac43d.png)

PR #1122, opened 2020-06-19
Fix for #1121 - flex the chips horizontally

blitzmann:
## **Description**
Fix for #1121 

Added styles to `.novo-chip-container` that flexes the chip components, allowing them to fall horizontally instead of stacking vertically

#### **Verify that...**

- [x] `npm run lint` passes
    * There are failures, but these are inherited from failures in `master`. This is a simple style update. 

- [ ] Run `Novo Automation`
- [ ] Run `BBO Automation`

### Before:
image

### After:
image

### DIFF
--- a/projects/novo-elements/src/elements/chips/Chips.scss
+++ b/projects/novo-elements/src/elements/chips/Chips.scss
@@ -72,6 +72,10 @@ novo-entity-chips {
       border: 1px solid $positive;
     }
   }
+  .novo-chip-container {
+    display: flex;
+    flex-wrap: wrap;
+  }
   .chip-input-container {
     flex-grow: 4;
     input {


tbo47:
Do you mind adding screenshots of before and after your changes?

blitzmann:
@tbo47 before/after images added

tbo47:
Thanks for the screenshots.

In our app we do:
```
.novo-chip-container {
    max-height: 24.5em;
    overflow: auto;
    display: flex;
    flex-wrap: wrap;
}
```

So your contribution is compatible with what we have. Let me run a couple of tests internally but it should be ok.

blitzmann:
Sounds good! And yeah, that's where we got difference between BH prod and what novo was giving. I would have thought the app would uses the same code, does it not? Anyway, let me know if any changes are required - we weren't sure if the max height and overflow properties were needed for novo or if they were app-specific :)

PR #1165, opened 2021-03-12
fix(Select): fixing logic around specialized lookup options

jhoac:
- Removing filter of readOnly on options so that it can contain all possible options (filter happens later in Select.ts to create filteredOptions)
- Changing writeValue to use more than just filteredOptions so that
  readOnly values can still render properly

## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**


projects/novo-elements/src/utils/form-utils/FormUtils.ts
```diff
@@ -691,7 +691,7 @@ export class FormUtils {
       field.dataSpecialization === 'SPECIALIZED_OPTIONS' ||
       (field.options && ['SpecializedOptionsLookup', 'SimplifiedOptionsLookup'].includes(field.dataType))
     ) {
-      return field.options.filter((o) => !o.readOnly);
+      return field.options;
```
monroepe, line 694: Could you add a unit test for this change if possible

monroepe (approved):

dvoegelin (approved):

PR #1171, opened 2021-03-25
F/app bridge timeout

hiqbal01:
## **Description**

Added an optional to timeout to app bridge http calls to enable calls that take longer and added a default timeout of 50s.

#### **Verify that...**

- [x] `npm run lint` passes
- [x] `npm run build` still work


bvkimball (commented):
I would rather keep the timeout default value at 10 seconds but all for this override

bvkimball (approved):

PR #1184, opened 2021-05-20
chore(actions): create manual action to publish novo elements 

charlesabarnes:
## **Description**

Create an action that publishes Novo elements to NPM automatically.  This is run manually from the actions tab.

### DIFF
--- a/.github/workflows/releaseNovoElements.yml
+++ b/.github/workflows/releaseNovoElements.yml
@@ -2,10 +2,10 @@ name: Release Novo Elements
 on:
   workflow_dispatch:
     inputs:
-    type:
-      description: 'Update Type (patch, major, minor)'
-      required: true
-      default: 'patch'
+      type:
+        description: 'Update Type (major, minor, patch)'
+        required: true
+        default: 'patch'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/releaseNovoElements.yml
```diff
@@ -0,0 +1,44 @@
+name: Release Novo Elements
+on:
+  workflow_dispatch:
+    inputs:
+    type:
+      description: 'Update Type (patch, major, minor)'
```
ndickerson, line 6: Not sure exactly how this gets consumed, but just for standardization, should this be:

`major, minor, patch`?

charlesabarnes: That bit is just a hint for for the options available in the action.   What is input gets used by `npm version` to bump the version number in the package.json and create the tag.   Updated the order, thank you

amrutha-rajiv (approved):

bvkimball (changes requested):
Don't we use standard-version to version novo-elements?

charlesabarnes:
> Don't we use standard-version to version novo-elements?
@bvkimball 

You're right, I simplified the script by just changing it to run `npm run release:elements`.  The postscript will handle publishing and everything

bvkimball (approved):

PR #1186, opened 2021-05-26
feat(DataTable): Expose sort and filter in novo-data-table

charlesabarnes:
## **Description**

Expose sort and filter as methods in novo-data-table.  This allows content on the table to be manipulated, without directly manipulating state.  The distinction for this feature versus manipulating `outsideFilter` is the persistence of the filter as a preference. 

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`
- [ ] Run `BBO Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/data-table/data-table.component.ts
+++ b/projects/novo-elements/src/elements/data-table/data-table.component.ts
@@ -260,7 +260,7 @@ export class NovoDataTable<T> implements AfterContentInit, OnDestroy {
   @HostBinding('class.global-search-hidden') globalSearchHiddenClassToggle: boolean = false;
 
   @ContentChildren(NovoTemplate) customTemplates: QueryList<NovoTemplate>;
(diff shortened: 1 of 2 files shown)


bvkimball:
Why is this better than invoking the action through the DataSource?

charlesabarnes:
> Why is this better than invoking the action through the DataSource?

Mostly it's about keeping track of the table's state.  

If you want to filter and leverage all of the data-table's features while also not using a separate column, it needs to go through the table state.  Instead of reinventing the wheel by re-implementing the sort and filter logic elsewhere, I think it's best to use the same logic from the `NovoDataTableSortFilter` directive.

We want stuff going through this method to be persisted to saved searches with the normal sort and filter data.

bvkimball:
I disagree i don't think you should go through the component to invoke this, it should happen through the DataSource... we should either expose state:DataTableState or add proxy methods to the DataSource IMO

charlesabarnes:
Marking as draft to reimplement my approach.

PR #1196, opened 2021-06-25
feat(dropdown): Add support for submenus in dropdowns

trenthibbard:
Previously, there was no way for a dropdown item to have a submenu. This enhancement adds submenuItems and onSubmenuClick as optional attributes on the NovoItemElement which can be used to create a menu within a menu.

## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`
##### **Screenshots**


bvkimball:
I have this psuedo supported in the `golden` branch....  i am worried about compatibility, since we are moving all "items" to "options"... we should sync on this.

trenthibbard:
Closing this as the functionality is no longer needed for what I was working on. If we want to have similar functionality in a future state, we can use this code for reference because this did work well, but merging this would complicate the larger refactor efforts

PR #1212, opened 2021-08-13
feat(tooltip): Adding HTML support for tooltips

sajjansarkar:
Other minor updates:
 - Fixed a broken unit test
 - Cleaned up the tool tip examples as the tooltip parents were squashed together

## **Description**

Adding an option to allow HTML tooltips

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**
image

### DIFF
--- a/projects/novo-elements/src/elements/form/Control.spec.ts
+++ b/projects/novo-elements/src/elements/form/Control.spec.ts
@@ -71,7 +71,7 @@ describe('Test Localization', () => {
 
   it('should set decimal separator based on locale correctly', () => {
     const component = new NovoControlElement(mockElement, null, null, null, null, null, 'fr-FR');
-    expect(component.decimalSeparator).toBe('.');
+    expect(component.decimalSeparator).toBe(',');
   });
 });
 
--- a/projects/novo-elements/src/elements/form/Control.spec.ts
+++ b/projects/novo-elements/src/elements/form/Control.spec.ts
@@ -71,7 +71,7 @@ describe('Test Localization', () => {
 
   it('should set decimal separator based on locale correctly', () => {
     const component = new NovoControlElement(mockElement, null, null, null, null, null, 'fr-FR');
-    expect(component.decimalSeparator).toBe(',');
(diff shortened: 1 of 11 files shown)


nickle799 (approved):

ndickerson (approved):

projects/novo-elements/src/elements/form/Control.spec.ts
```diff
@@ -71,7 +71,7 @@ describe('Test Localization', () => {
 
   it('should set decimal separator based on locale correctly', () => {
     const component = new NovoControlElement(mockElement, null, null, null, null, null, 'fr-FR');
-    expect(component.decimalSeparator).toBe('.');
+    expect(component.decimalSeparator).toBe(',');
```
ndickerson, line 74: I approve of the changes overall - I'm just not sure if this is going on here - is this a case where the test is broken on master, or just a branch? Is this a case where the decimal separator is actually being based on the real physical locale so that it works in some geographic locations but not others?

sajjansarkar: The unit test appears to be broken on master, i assumed the locale was being passed in as the last arg in line 73 and looks like France uses comma as a decimal separator vs a period.

ndickerson: Perfect. thanks!

nickle799 (approved):

bvkimball:
@ndickerson, @sajjansarkar  We have added this in the past several times, i don't think we should parse a string as html because it is unsafe, i also don't like the api for having isTooltipHTML.

I believe we landed on if this is HTML then it needs to be a template that is passed to the tooltip.

```
<ng-template #tooltip-content>
 <strong>Stop</strong> don't do this!
</ng-template>

<div [tooltip]="tooltip-content">Warning!</div>
```

This is sort of how POPOVERS work, so we had paused to determine the right path forward...

Either make it so that PopOvers can be "Dark" or Add Templates literals to tooltips.... maybe re-write both of them.

....

PR #1238, opened 2021-11-17
feat(chomsky): remove chomsky from novo-elements

charlesabarnes:
## **Description**

Removing Chomsky from novo-elements.  ngx-translate is the preferred replacement.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


bvkimball:
I would also make this change in `next`

PR #1244, opened 2021-12-06
feat(data table): Enable data table headers to word wrap

kiri-um-bh:
## **Description**
This fix allows data table headers to word wrap. Previously, the column titles were truncated if they were longer than the column width.


charlesabarnes (approved):
Discussed with UX and product and they gave the goahead

PR #1266, opened 2022-03-03
refactor(): some refactor work in preparation for typescript 4.0.8 upgrade

dvoegelin:
## **Description**

getting ready to update typescript from 3.9 to 4.0 in anticipation of angular 11 upgrade.
had to update a few components to use properties instead of getters/setters, and vice versa, due to breaking change in typescript 4.0.

#### **Verify that...**

- [/] Any related demos were added and `npm start` and `npm run build` still works
- [/] New demos work in `Safari`, `Chrome` and `Firefox`
- [/] `npm run lint` passes
- [/] `npm test` passes and code coverage is increased
- [/] `npm run build` still works

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "tslint": "~6.1.3",
     "tslint-config-prettier": "^1.14.0",
     "typedoc": "^0.20.36",
-    "typescript": "3.9.7",
+    "typescript": "4.0.8",
     "xmldom": "^0.6.0"
   },
   "jest": {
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "novo-elements",
-  "version": "6.0.0",
+  "version": "6.0.1",
```
charlesabarnes, line 3: This should be done by the release action

charlesabarnes (approved):

package.json
dvoegelin: reverted, thanks

PR #1287, opened 2022-04-08
fix(datepickerInput): fix calendar UI not respecting valid date in input field

JorgeBerroa:
## **Description**

I modified the _setCalendarValue method in NovoDatePickerInputElement class. The date picker input did not respect the date that was in the input field unless the enter key was pressed and then the calendar ui would highlight the date in the input field. The reason this was happening was that this code value = new Date(value).setHours(0, 0, 0, 0); was returning a date as a timestamp instead of a date object.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**


bvkimball (approved):

bvkimball:
@hiqbal01 can you or @dvoegelin  look at this for qa

PR #1288, opened 2022-04-08
fix(Security): providing security service in root to avoid running unnecessary instances

dvoegelin:
## **Description**

The Security service is being instantiated more times than we need it to be, so by providing it in root and removing it from the providers export and all but the main Novo module we are able to cut down the number of instances to one per iframe (plus the one in the mainframe).

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**


bvkimball (approved):

bvkimball:
We should evaluate if all of our services, i don't see why all wouldn't be `provideIn: root`

Issue #1267, opened 2022-03-09
Pagination Styling is not dynamic with differing backgrounds


worycatts:
_**Note: for support questions, please use one of these channels:**_ [Chat: BullhornOSS Gitter](https://gitter.im/bullhorn/Open-Source)

**Describe the bug**
Pagination does not have a dynamic background color indicating the page selection. Currently, #f4f4f4 is used as the background color indicating which page is selected. When pages themselves use the background color of #f4f4f4, the user cannot determine which page is selected due to the colors being the same. 

**To Reproduce**
1. Login with shifts enabled user - [Confluence User (Beta)](https://confluence.bullhorn.com/display/DEV/How+to+Test+Shifts+Code)
2. After logged in, open DevTools > Console and paste the following in the console
     a. debugMenu('https://shift-scheduler-f-real-golden.ets-dev2.kube.bullhorn.com/schedule')
3. When scheduler loads, select a different page in the header to view which page is selected

**Expected behavior**
Verify: The selected page is easily identifiable

**Actual behavior**
Actual: The user cannot determine which page is selected as the background pagination color is the same as the header background color  

**Specifications**
  - Version: v6.0.0
  - Platform:   Novo
  - Subsystem:

**Screenshots**
This screenshot shows that pagination is visible in Novo, but doesn't account for backgrounds with the same color. 
![image](https://user-images.githubusercontent.com/23424821/157515689-2679f8b7-3607-41ea-af65-0c191052d8d5.png)

Another example using the background colors of #f4f4f4
![image](https://user-images.githubusercontent.com/23424821/157516596-aac51105-d426-4180-812a-79b70f46a67b.png)

**Additional context**
This ticket is to create a dynamic background when selecting a page to easily identify which page the user is on regardless of the background color.

PR #1290, opened 2022-04-11
fix(NovoDataTable): making pagination marker background color dynamic

dvoegelin:
## **Description**

This resolves issue https://github.com/bullhorn/novo-elements/issues/1267 by making the background of the data table pagination marker dynamic. instead of a hardcoded color, it's now simply black with an opacity of 10%, so it will render as whatever color the header is, just slightly darker.

pagination background

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/data-table/data-table.component.scss
+++ b/projects/novo-elements/src/elements/data-table/data-table.component.scss
@@ -623,7 +623,7 @@ novo-data-table-pagination {
       }
       .page.active {
         color: $company;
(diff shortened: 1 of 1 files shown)


bvkimball (approved):

bvkimball:
I think this is just a Quick fix.  On darker backgrounds it probably should inverse, and font-text should update.

PR #1334, opened 2022-07-26
feat(NonIdealState): Non Ideal Loading

antonyw89:
## **Description**

Adding a message next to the novo-loading component is not a known pattern. This could be used to display a message while loading with the flag controlled at implementation. This can be used for a loading message displaying all the time for a loading screen, or having a timer flipping the flag to show a message during a long long loading screen.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**
image
image

### DIFF
--- a/projects/novo-examples/src/components/non-ideal-state/non-ideal-state-loading-usage/non-ideal-state-loading-usage-example.html
+++ b/projects/novo-examples/src/components/non-ideal-state/non-ideal-state-loading-usage/non-ideal-state-loading-usage-example.html
@@ -1,0 +1,7 @@
+<novo-non-ideal-state>
+  <novo-loading></novo-loading>
+  <novo-text lineLength="large" marginAfter>We are currently experiencing technical difficulites and your wait time is
+    taking a bit
+    longer than expected. Please be patient. </novo-text>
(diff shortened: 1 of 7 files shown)


bvkimball:
Looks great update the pr description

projects/novo-examples/src/components/non-ideal-state/non-ideal-state-loading-usage/non-ideal-state-loading-usage-example.html
```diff
@@ -0,0 +1,6 @@
+<novo-non-ideal-state>
+  <novo-loading></novo-loading>
+  <novo-text lineLength="large" marginAfter>We are currently experiencing technical difficulites and your wait time is
+    taking a bit longer than expected. Please be patient. </novo-text>
```
ndickerson, line 4: Super minor - I know this is just an example, but you can strike the second sentence. It's totally fine with just the first. The general guidance from UX we've received in the past is to never say Please to the user. Also, this makes it sound like they are impatient.

ndickerson (approved):

PR #1341, opened 2022-08-19
feat(DataTable): adding ability to explicitly set data-table state

dvoegelin:
..and reset saved search name

## **Description**

most of the data-table state fns are additive, allowing the consumer to change, set, reset one property without affecting the others. the exception to this is the reset() fn, which sets everything to undefined and then fires an update event. we needed a fn that functions like this, but sets everything to a provided value whether it's defined or not so we can explicitly set the state of the data-table.

we also needed a way to store and reset the name of the currently applied saved search to avoid having to manage this state separately in lower level services. 

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/data-table/data-table.component.ts
+++ b/projects/novo-elements/src/elements/data-table/data-table.component.ts
@@ -359,7 +359,7 @@ export class NovoDataTable<T> implements AfterContentInit, OnDestroy {
       // Re-subscribe
       this.outsideFilterSubscription = outsideFilter.subscribe((filter: any) => {
         this.state.outsideFilter = filter;
-        this.state.updates.next({ globalSearch: this.state.globalSearch, filter: this.state.filter, sort: this.state.sort, where: this.state.where, savedSearchName: this.state.savedSearchName });
+        this.state.updates.next({ globalSearch: this.state.globalSearch, filter: this.state.filter, sort: this.state.sort, where: this.state.where });
         this.ref.markForCheck();
       });
     }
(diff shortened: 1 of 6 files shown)


projects/novo-elements/src/elements/data-table/data-table.component.ts
```diff
@@ -358,7 +359,7 @@ export class NovoDataTable<T> implements AfterContentInit, OnDestroy {
       // Re-subscribe
       this.outsideFilterSubscription = outsideFilter.subscribe((filter: any) => {
         this.state.outsideFilter = filter;
-        this.state.updates.next({ globalSearch: this.state.globalSearch, filter: this.state.filter, sort: this.state.sort, where: this.state.where });
+        this.state.updates.next({ globalSearch: this.state.globalSearch, filter: this.state.filter, sort: this.state.sort, where: this.state.where, savedSearchName: this.state.savedSearchName });
```
bvkimball, line 362: Not sure the DataTable needs to know about the savedSearch???  Could this not be stored in the custom DataTableState service in your implmentation?

bvkimball (changes requested):

bvkimball:
With a few changes you probably could override `DataTableState` with you own `CustomDataTableState`

projects/novo-elements/src/elements/data-table/data-table.component.ts
dvoegelin: i pulled it out of all the `updates` events and the state clear functions. i kept it in the new setState function since that's needed to set the preferences and the preferencesChanged event emitter since that's passing an IDataTablePreferences obj which includes the savedSearchName.

hiqbal01 (approved):

bvkimball (approved):

PR #1352, opened 2022-09-15
feat(dropdown): making dropdown text selectable

antonyw89:
## **Description**

Updating dropdown button text to be selectable again.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**
image

### DIFF
--- a/projects/novo-elements/src/elements/dropdown/Dropdown.scss
+++ b/projects/novo-elements/src/elements/dropdown/Dropdown.scss
@@ -12,7 +12,7 @@ novo-dropdown {
   novo-button {
     position: relative;
     z-index: 0;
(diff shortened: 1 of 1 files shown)


bvkimball:
Why? I don't think buttons should have selectable text?

antonyw89:
> Why? I don't think buttons should have selectable text?

There has been requests from users to make dropdown text selectable again. They were previously selectable before the novo UI updates.

bvkimball:
Bug is caused by Buttons that are styled to look like "Select Inputs"... This code is deprecated and will be removed shortly, suggested solution is to change the implementation to use a "novo-select" input.

PR #1353, opened 2022-09-20
fix(Chips): BH-77703 updates chips to show match object

kylearandall:
## **Description**

Reverted Chips.ts match object to contain the full selected object instead of just the value. 
Needed to fix this bug: https://jira.bullhorn.com/browse/BH-77703
Novo code where bug is located: https://bhsource.bullhorn.com/NOVO/novo/-/blob/development/libs/bte/src/components/rule/add-edit-rule-form/holiday-picker-preview/holiday-picker-preview.ts#L30

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**
Screenshot 2022-09-20 153642

### DIFF
--- a/projects/novo-elements/src/elements/chips/Chips.ts
+++ b/projects/novo-elements/src/elements/chips/Chips.ts
@@ -311,7 +311,7 @@ export class NovoChipsElement implements OnInit, ControlValueAccessor {
       if (!this.popup) {
         this.popup = this.componentUtils.append(this.source.previewTemplate, this.preview);
       }
-      this.popup.instance.match = { data: this.selected.data ?? this.selected.value };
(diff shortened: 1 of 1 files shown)


bvkimball:
I believe this would break the default Preview Template which is EntityPickerResult.  PreviewTemplates, Pickers, & Chips always had an optional value format due to typing things `any`.

I am confused why your picker results doesn't add the object to the "data" property.  

The following might be more right... but i would rather keep it tied to `value`.

```
      this.popup.instance.match = { data: this.selected.data ?? this.selected};
```

In Novo can you try to get your options in the following format `{value, label, data}`... or we can make the change above, but either way your picker preview needs to expect a `{data: object}`

PR #1382, opened 2023-01-03
feat(NovoDataTable): updating data table logic and template to better facilitate custom filters

dvoegelin:
## **Description**

I ran into a few issues when implementing a custom filter in the Bullhorn ATS, so I made some changes to the NovoDataTable and related components to better accommodate custom filters and their styling and functionality.

- the filter header was not customizable, as it was a static component outside of the custom template container. this was pulled out into its own component so it can be used anywhere, while still maintaining it in the original position if column.filterable.useCustomHeader is not set to true. this was done to make this change non-breaking.
- the custom filter was attributed to a column based on the column.id property, so this meant that every column needed to have its own custom template, instead of being able to use one custom template and attribute it to multiple columns based on, for example, the column's type. this was updated to key off of the new optional column.filterable.customTemplate property, but will still fall back to use column.id if this property is not set.
- these two new properties have been added to the IDataTableColumnFilterConfig interface as optional.
- max width on the data-table-dropdown class has been upped from 230px to 280px
- the custom filter template was previously embedded within a novo-option, which applied immutable and potentially undesirable styling properties to the custom filter. this was considered a bug and fixed by instead embedding it in a simple ng-container which does not apply any styling to the custom filter. technically this will remove 1rem of padding from around any existing custom filters, so could be considered a breaking change.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


ndickerson (approved):

projects/novo-elements/src/elements/data-table/cell-headers/data-table-header-cell-filter-header.component.ts
```diff
@@ -0,0 +1,42 @@
...
+        *ngIf="hasFilter"
+        data-automation-id="novo-data-table-filter-clear">
+        {{ labels.clear }}
+      </novo-button>
+    </div>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class NovoDataTableCellFilterHeader {
+  @Input() label: any;
```
monroepe, line 25: could this be string instead of any? Might be safer to keep it as any in case its like an int or something

dvoegelin: yeah i wanted to keep it open so it can be used however people want, but really it should never be anything other than a string or an int. so that's a good call. made this a bit stricter.

monroepe (approved):

ndickerson (approved):

PR #1410, opened 2023-04-12
BH-77510 update css to fix overflow issue

JorgeBerroa:
## **Description**

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/dropdown/Dropdown.scss
+++ b/projects/novo-elements/src/elements/dropdown/Dropdown.scss
@@ -30,6 +30,8 @@ novo-dropdown {
   margin-top: $spacing-sm;
   margin-bottom: $spacing-sm;
   box-shadow: $shadow-3;
+  max-height: calc(100vh - 400px);
+  overflow-y: scroll;
   &.has-height {
     overflow: auto;
   }


projects/novo-elements/src/elements/dropdown/Dropdown.scss
```diff
@@ -30,6 +30,9 @@ novo-dropdown {
   margin-top: $spacing-sm;
   margin-bottom: $spacing-sm;
   box-shadow: $shadow-3;
+  max-height: calc(100vh - 400px);
+  overflow-y: scroll;
```
davidkbh, line 34: Do we want overflow-y scroll, or auto? Do we always want the scrollbar to appear, even for just a few options?

JorgeBerroa: I used scroll because that's the same design our actions dropdown menu on entities uses. I do not mind changing this to auto but I think the style design be coherent for the dropdown.

davidkbh (approved):

projects/novo-elements/src/elements/dropdown/Dropdown.scss
dvoegelin: dropdown styling changes

was this tested in the demo? i agree with david here that we should not be displaying the scroll bar if there are not enough options to warrant the scroll bar. the min-height property is also creating empty space at the bottom of the dropdown if there are less than 6 or 7 options. we need to rethink the solution for this.

dvoegelin: also please add a description to the PR, and unless this is being hotfixed it should be going into the `next` branch, not `master`.

PR #1415, opened 2023-04-20
feat(EntityColor): Add compliance manager entity color

ZacharyJorn:
## **Description**

Add compliance manager entity color and update icons.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/styles/variables.scss
+++ b/projects/novo-elements/src/styles/variables.scss
@@ -123,6 +123,7 @@ $entity-colors: (
   "earnCode": $color-earn-code,
   "billableCharge": $color-billable-charge,
   "payableCharge": $color-payable-charge,
+  "complianceManager": $color-positive,
   "invoiceStatement": $color-invoice-statement,
 );
 // Generic Color Map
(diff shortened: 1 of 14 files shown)


dvoegelin (approved):

projects/novo-elements/src/styles/variables.scss
```diff
@@ -123,6 +123,7 @@ $entity-colors: (
   "earnCode": $color-earn-code,
   "billableCharge": $color-billable-charge,
   "payableCharge": $color-payable-charge,
+  "complianceManager": $color-positive,
```
ndickerson, line 126: I think you need to follow the existing pattern of entity colors like invoiceStatement and include this in all three locations.

ZacharyJorn: This isnt technically an entity though. Also there are other things in that list that are only in that list and in no other lists. We only need it there to show the correct background color for a menu item

ndickerson (approved):

PR #1416, opened 2023-04-21
WIP: fix(Chips): Prevent overwriting value before labels are resolved

trenthibbard:
When altering the value of a Chips picker, if some values have a label while others do not, the items with no label are no longer removed from the value

## **Description**

Previously, an async issue in Chips setItems() method would result in removing items from the value if there was a mix of items with labels and items without labels. This pull request will remedy this by only setting the value for the Chips element at the appropriate time

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


projects/novo-elements/src/elements/chips/Chips.ts
```diff
@@ -175,15 +175,13 @@ export class NovoChipsElement implements OnInit, ControlValueAccessor {
               this.items.push(value);
             }
           }
-          this._items.next(this.items);
+          this._finalizeItemValue();
```
monroepe, line 178: it seems like we want to call `this._finalizeItemValue();` in every scenario or close to every scenario. Could we move this call to the last line of the function (and remove the other calls to it) so it gets call every time? The logic is a little hard to follow with all the if/elses so I understand if that is not possible. I think at least in this part of the function we could move it outside of this if check.

alyssabryant: thanks for the suggestion! I will look at getting rid of the else blocks and just ending the method with this

alyssabryant: was able to simplify this a little bit, we do still need to call this new method within the `then` block for the 'noLabels' scenario in order to resolve the issue

monroepe (approved):

dvoegelin (approved):

PR #1437, opened 2023-06-29
fix(Popover): renaming popover attribute due to new HTML spec attribute

dvoegelin:
## **Description**

the HTML spec just updated to now include its own `popover` attribute, overriding our novo-elements directive with the same name in certain implementations, rendering them broken. this is manifesting itself in the element being hidden with a console warning informing you that it "Found a 'popover' attribute with an invalid value.". the fix for this is to rename our attribute `novoPopover`, while still retaining the old name `popover` to retain backwards compatibility for anyone using them with a non-broken implementation.

any app that implements a novo-elements popover without bracketing the directive will need to update their implementation to use the new naming convention. for example:
`<span popover="popover text"></span>`  is broken. the novo-elements directive in this case will be overridden by the new attribute.
`<span [popover]="popoverTextProp"></span>` is not broken, the directive in this case is not overridden.
if you are using the former implementation, this PR will allow you to simply update the name to `novoPopover`.

https://html.spec.whatwg.org/multipage/popover.html

this broke with latest chrome update. as other browsers update to support the new attribute they will stop working as well. firefox is still working as of v114. current list of browser compatibility with the new attribute can be seen here [Popover API - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API#browser_compatibility)

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


projects/novo-elements/src/elements/form/FieldInteractionApi.ts
```diff
@@ -495,7 +495,7 @@ export class FieldInteractionApi {
       control.popoverDisabled = popover.disabled;
       control.popoverAnimation = popover.animation;
       control.popoverDismissTimeout = popover.dismissTimeout;
-      this.triggerEvent({ controlKey: key, prop: 'popover', value: popover }, otherForm);
+      this.triggerEvent({ controlKey: key, prop: 'novoPopover', value: popover }, otherForm);
```
ndickerson, line 498: You can leave this as the normal "popover" word

monroepe (approved):

ndickerson (approved):

PR #1456, opened 2023-08-14
fix(Dropdown): fixing a styling bug with data table dropdown cell lists

dvoegelin:
## **Description**

certain DataTable cell dropdowns were relying on styles located in a stylesheet for the Table component, so after we went through and encapsulated styles properly they were no longer getting applied. The Table component is deprecated so the best fix for this was just to copy these styles from that component stylesheet and put them in the appropriate Dropdown component stylesheet.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/dropdown/Dropdown.scss
+++ b/projects/novo-elements/src/elements/dropdown/Dropdown.scss
@@ -49,4 +49,29 @@
       background: $light;
     }
   }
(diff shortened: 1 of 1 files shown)


dvoegelin:
point fixing instead, closed in favor of #1457

PR #1479, opened 2023-11-09
docs(DragDrop): Updated drag-drop example

davidkbh:
## **Description**

Updates to the drag drop example to make it more presentable.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**
dragdrop-newdemo

### DIFF
--- a/projects/novo-examples/src/utils/drag-drop/drag-drop/drag-drop-example.html
+++ b/projects/novo-examples/src/utils/drag-drop/drag-drop/drag-drop-example.html
@@ -1,19 +1,32 @@
+
 <p>Click and drag these buttons to place them in a new position. When rearranged, they emit an event to demonstrate the new order.</p>
 <div [novoDragDrop]="objects" (novoDragDropFinish)="dragFinished($event)" class="drag-box">
-    <div *ngFor="let item of objects"
+    <novo-card *ngFor="let item of objects"
         class="demo-drag-item">
-        <button class="novo-drag-target" [style.backgroundColor]="item.bgColor">{{item.btnText}}</button> <button class="remove-btn" (click)="removeObject(item)">X</button>
-        <div novoPreventDrag>{{item.name}} (Non-draggable area)</div>
-    </div>
+        <novo-card-header class="novo-drag-target {{item.bgClass}}">
+            <novo-title>{{item.headerText}}</novo-title>
+            <novo-action (click)="removeObject(item)" icon="close" tooltip="Close Item"></novo-action>
+        </novo-card-header>
+        <novo-card-content>
+            <novo-caption novoPreventDrag>{{item.name}} (Non-draggable area)</novo-caption>
+        </novo-card-content>
+    </novo-card>
 </div>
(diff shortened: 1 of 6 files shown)


projects/novo-examples/src/utils/drag-drop/drag-drop/drag-drop-example.ts
```diff
@@ -17,23 +17,23 @@ interface DemoButtonObject {
...
 export class DragDropExample implements OnChanges {
   objects: DemoButtonObject[] = [
-    { name: 'Object 1', btnText: 'Button 1', bgColor: 'red' },
-    { name: 'Object 2', btnText: 'Button 2', bgColor: 'blue' },
-    { name: 'Object 3', btnText: 'Button 3', bgColor: 'green' },
-    { name: 'Object 4', btnText: 'Button 4', bgColor: 'yellow' },
-    { name: 'Object 5', btnText: 'Button 5', bgColor: 'wheat' },
-    { name: 'Object 6', btnText: 'Button 6', bgColor: 'purple' },
-    { name: 'Object 7', btnText: 'Button 7', bgColor: 'teal' }
+    { name: 'Object 1', headerText: 'Item 1', bgColor: 'rgba(255,0,0,0.3)' },
```
ndickerson, line 20: Would it make sense to use the standard bullhorn colors here?

davidkbh: At first, I was preferring something more muted to make sure they'd all be readable with black text. This came up in standup though, and since the word "Title" is hardly important, I think those do work out better. I've just updated it.

projects/novo-examples/src/utils/drag-drop/drag-drop/drag-drop-example.html
```diff
@@ -1,19 +1,32 @@
...
 <div [novoDragDrop]="objects2" (novoDragDropFinish)="dragFinished2($event)" class="drag-box">
-    <div *ngFor="let item of objects2"
+    <novo-card *ngFor="let item of objects2"
         class="demo-drag-item">
-        <button class="novo-drag-target" [style.backgroundColor]="item.bgColor">{{item.btnText}}</button> <button class="remove-btn" (click)="removeObject(item)">X</button>
-        <div novoPreventDrag>{{item.name}} (Non-draggable area)</div>
-    </div>
+        <novo-card-header class="novo-drag-target {{item.bgClass}}">
+            <novo-title>{{item.headerText}}</novo-title>
+            <novo-action (click)="removeObject(item)" icon="close" tooltip="Close Item"></novo-action>
```
dvoegelin, line 26: this is calling the same removeObject function from as the demo above, which only removes items from the `objects` array populating that demo. might want to update this.

davidkbh: Ah, you're right - tried this on the first list but not the second. Thanks for spotting.

dvoegelin (approved):

monroepe (approved):

PR #1498, opened 2024-01-17
feat(AppBridge): When handling AppBridge events, pass origin to handler

davidkbh:
## **Description**

This change takes the raw "origin" value from postrobot event handlers and appends it to the message data, making it easier for the host frame to record information on different senders. This comes in natively as part of the iframe message event, so the child instance of AppBridge doesn't need to implement it.

This also changes the types around MessageTypes so that they are definitively typed towards each string, rather than all being of type "string". Thus, a coder providing a value that doesn't match the options will get a compilation error.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/utils/app-bridge/AppBridge.ts
+++ b/projects/novo-elements/src/utils/app-bridge/AppBridge.ts
@@ -80,32 +80,45 @@ export interface IAppBridgeRequestDataEvent {
   type: NovoDataType;
 }
 
-const HTTP_VERBS = {
-  GET: 'get',
-  POST: 'post',
-  PUT: 'put',
-  DELETE: 'delete',
+namespace HTTP_VERBS {
+  export const GET = 'get';
+  export const POST = 'post';
+  export const PUT = 'put';
+  export const DELETE = 'delete';
 };
 
-const MESSAGE_TYPES = {
-  REGISTER: 'register',
-  OPEN: 'open',
-  OPEN_LIST: 'openList',
-  CLOSE: 'close',
-  REFRESH: 'refresh',
-  PIN: 'pin',
-  PING: 'ping',
-  UPDATE: 'update',
-  HTTP_GET: 'httpGET',
-  HTTP_POST: 'httpPOST',
-  HTTP_PUT: 'httpPUT',
-  HTTP_DELETE: 'httpDELETE',
-  CUSTOM_EVENT: 'customEvent',
-  REQUEST_DATA: 'requestData',
-  CALLBACK: 'callback',
+namespace MESSAGE_TYPES {
+  export const REGISTER = 'register';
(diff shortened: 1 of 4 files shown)


projects/novo-elements/src/utils/app-bridge/AppBridge.ts
```diff
@@ -151,129 +164,160 @@ export class AppBridge {
...
+    msgType: MessageType,
+    handler: AppBridgeHandler,
+    packet: T,
+    echoPacket: any,
+    resolveEventData: (any) => boolean,
+  }): Promise<boolean> {
+    let returnPromise: Promise<any>;
+    if (this._handlers[handler]) {
+      // Should be directly returning a promise. However, as a fallback, provide callback arguments
+      let callbackS, callbackF;
```
ndickerson, line 301: Is there a better name for these than s and f?

projects/novo-elements/src/utils/app-bridge/AppBridge.ts
```diff
@@ -151,129 +164,160 @@ export class AppBridge {
...
+      // Custom Events
+      [MESSAGE_TYPES.CUSTOM_EVENT]: async (event: PostRobotEvent<any>) => {
+        if (this._eventListeners[event.data.event]) {
+          this._eventListeners[event.data.event].forEach((listener) => {
+            listener(event.data.data);
+          });
+        }
+        if (this._registeredFrames.length > 0) {
+          this._registeredFrames.forEach((frame) => {
+            // TODO: Should this make sure it doesn't echo the custom event back to the author?
```
ndickerson, line 270: Would be nice to answer this question in code review before final merge. What are the pros/cons here?

davidkbh: So the way I picture this is, you have three frames in a tree; A is the parent of B and C. B emits a custom event, and having no specific idea of its needs, A emits it to both B and C.

Most of the time, B probably wouldn't have code responding to this custom event. But, if B and C are the same type of frame, it's possible for them to get confused by receiving an event off of their own action. eg, if we have a "FileAddedToProfile" event, B may have already added that file itself, fired the event, and then add it a second time from seeing the event come back.

ndickerson: Ahh, I see. Echoing back to the author is the easiest logic, but might not be the ideal solution. I think we can safely keep it as-is, and publishers who also want to consume the same custom events can ignore their own within their own logic.

ndickerson (approved):

dvoegelin (approved):

ndickerson (approved):

dvoegelin (approved):

PR #1509, opened 2024-02-16
feat(AppBridge): When handling AppBridge events, pass origin to handler

davidkbh:
## **Description**
This change takes the raw "origin" value from postrobot event handlers and appends it to the message data, making it easier for the host frame to record information on different senders. This comes in natively as part of the iframe message event, so the child instance of AppBridge doesn't need to implement it.

This also changes the types around MessageTypes so that they are definitively typed towards each string, rather than all being of type "string". Thus, a coder providing a value that doesn't match the options will get a compilation error.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


projects/novo-elements/src/utils/app-bridge/AppBridge.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+    it('should register a new app', async () => {
+        handleFn.mockReturnValue('terry');
+        await frame1Bridge.register();
+        expect(handleFn).toHaveBeenCalledWith(AppBridgeHandler.REGISTER, jasmine.objectContaining({ id: 'childapp1' }));
+        expect(hostBridge['_registeredFrames'].length).toBe(1);
+        expect(frame1Bridge.windowName).toBe('terry');
+    });
+
+    it('opens a window', async () => {
+        // TODO
```
monroepe, line 111: is there any way we could write these unit tests instead of leaving them as TODOs? In general I am not a fan of having TODOs in the code base and I think it would be good to have unit test coverage

ndickerson (approved):

ndickerson:
Nice work adding all of the new types and tests!

monroepe (approved):

projects/novo-elements/src/utils/app-bridge/AppBridge.spec.ts
davidkbh: Talked over this locally. While these are good tests to add, I'm not sure they're a priority just yet.

PR #1525, opened 2024-03-26
fix(Chips): hidden-chips, existing model data

kflo411:
## **Description**

- fixed edge-case (not found in examples) where model data is passed in to Chips.ts and hiddenChipsLimit is enabled/used. 

#### **Verify that...**

- [x] `npm run lint` passes

##### **Screenshots**


dvoegelin (dismissed):

dvoegelin:
if this issue wasn't caught to begin with because it was not accounted for in any of the demos, it may help to add this case to one of them with this PR.

amrutha-rajiv (approved):

PR #1531, opened 2024-04-11
fix(Progress Bar): Check for a valid total to avoid wrong percentages

Vonterio:
## **Description**

Add check for the progress total. If greater than 0, proceed as normal. If not then set percentage to be 0 to avoid NaN

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**


dvoegelin (approved):

projects/novo-elements/src/elements/progress/ProgressBar.ts
```diff
@@ -97,7 +97,12 @@ export class NovoProgressBarElement implements ControlValueAccessor, OnInit {
     if (this.value !== value) {
       this._value = value;
       if (this.progress) {
-        this._percent = this.value / this.progress.total;
+        if (this.progress.total > 0) {
+          this._percent = this.value / this.progress.total;
+        } else {
+          this._percent = 0;
```
ndickerson, line 103: If these changes are the exact same change in two places, we could either take the time to make this a single method, so that there's no duplication (Don't repeat yourself principle) or you could keep these changes to a single line, in which you don't need a method. 

```
this._percent = this.progress.total > 0 ? this._value / this.progress.total : 0;
```

Vonterio: Very true, fixed this

ndickerson (approved):

dvoegelin (approved):

davidkbh (approved):

PR #1556, opened 2024-06-17
feat(FieldInteractionApi): Make addControl work for novo control groups

robek2020:
## **Description**

Added a handler method to make the field interaction api addControl method work for novo form groups.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


robek2020:
We expected to need to use this functionality for some custom work in BH, and it does have some potential use cases (removing and adding controls per row in a control group might be good someday) but it's no longer necessary and doesn't have any immediate use. Closed.

PR #1561, opened 2024-06-21
fix(CKEditor): disabling version check to avoid licensing notification

dvoegelin:
## **Description**
ck editor is going to be displaying a notification/warning on all instances using version 22 and earlier telling users to update to the new licensed version, so this config update is to disable that warning.

https://ckeditor.com/blog/important-update-for-ckeditor-4-users/?utm_campaign=cke4_security_patch_lts&utm_source=pardot&utm_medium=email&utm_term=june-preannouncement&utm_content=esm

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


dvoegelin:
turns out this wasn't needed since we use an npm fork

PR #1563, opened 2024-06-28
feat(Data Table): Allow skipping explicit table.state update

kflo411:
## **Description**

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/data-table/data-table-clear-button.component.ts
+++ b/projects/novo-elements/src/elements/data-table/data-table-clear-button.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, ChangeDetectorRef, Component, EventEmitter, Output } from '@angular/core';
+import { ChangeDetectionStrategy, ChangeDetectorRef, Component, EventEmitter, Input, Output } from '@angular/core';
 import { NovoLabelService } from 'novo-elements/services';
 import { DataTableState } from './state/data-table-state.service';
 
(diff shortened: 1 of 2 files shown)


projects/novo-elements/src/elements/data-table/data-table-clear-button.component.ts
```diff
@@ -44,6 +44,8 @@ export class NovoDataTableClearButton<T> {
   queryClear: EventEmitter<boolean> = new EventEmitter();
   @Output()
   allClear: EventEmitter<boolean> = new EventEmitter();
+  @Input()
+  skipUpdate: boolean;
```
dvoegelin, line 48: we might only be concerned about the clearSearch function for the work we're doing now, but this input will be confusing if it's only applied to that one function and none of the others.

projects/novo-elements/src/elements/data-table/data-table-clear-button.component.ts
```diff
@@ -58,8 +60,13 @@ export class NovoDataTableClearButton<T> {
   }
 
   clearSearch(): void {
-    this.state.clearQuery();
-    this.queryClear.emit(true);
+    if (this.skipUpdate) {
+      this.queryClear.emit(true);
+      this.state.clearQuery(false);
```
dvoegelin, line 65: unless the ordering of these two calls matters, we can probably get rid of the if/else statement and just pass `!this.skipUpdate` into this function.

PR #1566, opened 2024-07-02
test(TabbedGroupPicker): Change the automation-ids on options to expr…

davidkbh:
…ess value rather than label

## **Description**

No changes to functionality. This replaces the data-automation-id on TabbedGroupPicker options so it presents the field's "name" rather than "label". eg:

label: "Favorite Color"
name: "customText5"

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/tabbed-group-picker/TabbedGroupPicker.html
+++ b/projects/novo-elements/src/elements/tabbed-group-picker/TabbedGroupPicker.html
@@ -48,7 +48,7 @@
           #tabbedGroupPickerVirtualScrollViewport>
           <novo-option
             *cdkVirtualFor="let item of displayTab.data"
-            [attr.data-automation-id]="item[displayTab.labelField]"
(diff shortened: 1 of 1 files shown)


projects/novo-elements/src/elements/tabbed-group-picker/TabbedGroupPicker.html
```diff
@@ -48,7 +48,7 @@
           #tabbedGroupPickerVirtualScrollViewport>
           <novo-option
             *cdkVirtualFor="let item of displayTab.data"
-            [attr.data-automation-id]="item[displayTab.labelField]"
```
dvoegelin, line 51: technically this is a breaking change, but i don't think anyone uses these automation-ids besides us so i'm okay with it. if we want to be 100% safe we can leave this as-is and add the valueField as a `data-automation-value`, but that will probably just be more confusing for people in the future.

PR #1591, opened 2024-08-21
fix(types): Update keyword types

ndickerson:
## **Description**

Updating keyword types in the data table to account for both basic keywords (list of strings) and expanded keywords (list of complex keyword objects). The functionality will apply to all future lists going forward that are connected to expanded keywords, which are also saved into preferences and saved searches. 

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/data-table/interfaces.ts
+++ b/projects/novo-elements/src/elements/data-table/interfaces.ts
@@ -20,7 +20,7 @@ export interface IDataTablePreferences {
 export interface DataTableWhere {
   query: string;
   criteria?: AdaptiveCriteria;
-  keywords?: string[];
+  keywords?: SearchKeywords;
   form: any;
 }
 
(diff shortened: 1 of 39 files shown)


projects/novo-elements/src/elements/data-table/interfaces.ts
```diff
@@ -20,7 +20,7 @@ export interface IDataTablePreferences {
 export interface DataTableWhere {
   query: string;
   criteria?: AdaptiveCriteria;
-  keywords?: string[];
+  keywords?: SearchKeywords;
```
dvoegelin, line 23: i don't think this update is going to be backwards compatible.

dvoegelin: this type was added recently enough that we're confident no one else is using it, so we're going to move forward with this update anyway.

dvoegelin (approved):

kflo411 (approved):

PR #1614, opened 2024-10-02
fix(Dependencies): overriding bugged rollup version

dvoegelin:
## **Description**

there appears to be a bug in a recent version of rollup that's causing local build failures. a clean npm install with the latest version works, so it seems to be an incompatibility with a previous version of one of our other dependencies. rolling rollup back to 4.11.0 for now and will update all of our dependencies and remove this override with the next minor release.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4258,8 +4258,178 @@
         "url": "https://github.com/sponsors/jonschlinkert"
       }
     },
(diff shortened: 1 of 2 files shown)


dvoegelin:
going to close this, moving forward with the workaround to simply delete node_modules and re npm install.

PR #1618, opened 2024-10-09
feat(Autocomplete): Clear the input field when the user loses focus

davidkbh:
## **Description**

The autocomplete input was appearing to clear out when users clicked away, but was just turning invisible. We've decided to actually clear the value, which still makes clear to the user that the text entered is not a full "value" (still needs a separator to turn it into a chip)

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/chips/ChipInput.ts
+++ b/projects/novo-elements/src/elements/chips/ChipInput.ts
@@ -127,10 +127,10 @@ export class NovoChipInput implements NovoChipTextControl, OnChanges, OnDestroy 
   }
 
   /** Checks to see if the blur should emit the (chipEnd) event. */
-  _blur(blurEvt: FocusEvent) {
+  _blur(blurEvent: FocusEvent) {
     if (this.addOnBlur) {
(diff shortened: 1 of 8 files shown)


projects/novo-elements/src/elements/chips/ChipInput.ts
```diff
@@ -117,9 +127,11 @@ export class NovoChipInput implements NovoChipTextControl, OnChanges {
   }
 
   /** Checks to see if the blur should emit the (chipEnd) event. */
-  _blur() {
+  _blur(blurEvt: FocusEvent) {
```
monroepe, line 130: its ok to keep it as-is. But i feel like naming the param `blurEvent` would have been a little more clear for only 2 extra characters

monroepe (dismissed):

kflo411 (approved):

monroepe (approved):

kflo411 (approved):

PR #1622, opened 2024-10-24
feat(Tooltip): tooltipOnOverflow, proper reactive resizing enablement

kflo411:
## **Description**

Implemented properly reactive tooltip enablement upon width overflow

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/tooltip/Tooltip.directive.ts
+++ b/projects/novo-elements/src/elements/tooltip/Tooltip.directive.ts
@@ -4,6 +4,7 @@ import { ComponentPortal } from '@angular/cdk/portal';
 import { AfterViewInit, Directive, ElementRef, HostListener, Input, OnDestroy, OnInit, ViewContainerRef } from '@angular/core';
 // APP
 import { NovoTooltip } from './Tooltip.component';
+import { BooleanInput } from 'novo-elements/utils';
 
 @Directive({
   selector: '[tooltip]',
@@ -40,8 +41,9 @@ export class TooltipDirective implements OnDestroy, OnInit, AfterViewInit {
   isHTML: boolean;
   @Input('tooltipCloseOnClick')
   closeOnClick: boolean = false;
+  @BooleanInput()
   @Input('tooltipOnOverflow')
-  overflow: boolean = false;
+  onOverflow: boolean = false;
 
   private tooltipInstance: NovoTooltip | null;
   private portal: ComponentPortal<NovoTooltip>;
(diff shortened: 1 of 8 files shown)


projects/novo-elements/src/elements/tooltip/Tooltip.directive.ts
```diff
@@ -40,10 +40,13 @@ export class TooltipDirective implements OnDestroy, OnInit {
   isHTML: boolean;
   @Input('tooltipCloseOnClick')
   closeOnClick: boolean = false;
+  @Input('tooltipOnOverflow')
+  overflow: boolean = false;
```
ndickerson, line 44: I think it would be easier to read if the variable was also called: `tooltipOnOverflow`, just because the word `overflow` is quite overloaded and it's harder for me to distinguish what the intent is later on in the file.

ndickerson, line 44: Maybe to be consistent, we could call the variable: `onOverflow` just because the word `overflow` is quite overloaded.

projects/novo-elements/src/elements/tooltip/Tooltip.directive.ts
```diff
@@ -87,7 +90,18 @@ export class TooltipDirective implements OnDestroy, OnInit {
     }
   }
 
+  ngAfterViewInit(): void {
+    if (this.overflow && this.elementRef?.nativeElement) {
+      this._resizeObserver = new ResizeObserver(() => {
+        const isOverflowing = this.elementRef.nativeElement.scrollWidth > this.elementRef.nativeElement.clientWidth;
```
davidkbh, line 96: Do you run into any issues with this around change detection where it doesn't re-render based on `active`? You might need to wrap the callback code within: `this.zone.run(() =>`

https://dev.to/christiankohler/how-to-use-resizeobserver-with-angular-9l5

davidkbh (approved):

ndickerson (approved):

ndickerson (dismissed):

davidkbh (dismissed):

davidkbh:
Currently, my branch (based on this one) is running into an error with ResizeObserver being unavailable in the test environment. You may run into the same thing when trying to rebase this with `next`. We could polyfill it with something like this:
https://stackoverflow.com/a/65606342/576396

kflo411:
> Currently, my branch (based on this one) is running into an error with ResizeObserver being unavailable in the test environment. You may run into the same thing when trying to rebase this with `next`. We could polyfill it with something like this: https://stackoverflow.com/a/65606342/576396

@davidkbh , I saw this as well and fixed it with a global jest mock in our setupJest.ts file.
See commit here for BH-90588: https://bhsource.bullhorn.com/NOVO/novo/-/merge_requests/5629/diffs?commit_id=ac4250a5b5b31d0db8fa73275fb098270f9b9564

davidkbh (approved):

ndickerson (approved):

PR #1631, opened 2024-11-04
feat(Condition Builder): Add reset method to condition builder

vladyslavsemenets:
## **Description**

I added a reset method that takes a values object where key is the key of the form control and value is the value that should be set for the form control.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/query-builder/condition-builder/condition-builder.component.ts
+++ b/projects/novo-elements/src/elements/query-builder/condition-builder/condition-builder.component.ts
@@ -187,15 +187,20 @@ export class ConditionBuilderComponent implements OnInit, OnChanges, AfterConten
     this.results$ = Promise.resolve(this.fieldConfig.options);
   }
 
-  reset(values: Record<string, any>): void {
-    // Reset form controls based on the values provided.
-    Object.entries(values).forEach(([key, value]) => {
-      this.parentForm.get(key).setValue(value, { emitEvent: false, onlySelf: true });
-    });
-    // Reinitialize the field and operator template.
+  /**
+   * Resets the input and operator view containers, regenerates the field templates,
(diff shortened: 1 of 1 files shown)


projects/novo-elements/src/elements/query-builder/condition-builder/condition-builder.component.ts
```diff
@@ -187,6 +187,17 @@ export class ConditionBuilderComponent implements OnInit, OnChanges, AfterConten
     this.results$ = Promise.resolve(this.fieldConfig.options);
   }
 
+  reset(values: Record<string, any>): void {
```
davidkbh, line 190: This seems like a good way of making the value changes, only critiques:

- We might want a different function name. My suggestion for `reset` was under the assumption it was only ever going to clear out the existing values, and return to the default operator with null value. If it's taking values in, then maybe it needs a different name like `setFormValues`.
- If this takes a `values` argument, someone may pass in the `field` key. Have you tested what happens if that is modified as well? (Depending on the component's configuration, the field selector may or may not be hidden)

I am assuming that for the purposes of a cancelable form, you'll be calling this with something like `conditionBuilder.reset({ operator: null, value: null })` when the user clicks Cancel and no previous values have been saved?

vladyslavsemenets: I think we can reset form values ​​from parent component, no need to do it here, initially I encountered a problem that after I reset the values ​​from the code, the operator template and fields remained the same, because of this I came up with a hack to hide/show the component, so I think now a function that simply resets these templates and initializes them again will be enough, and I will implement the logic of resetting the values ​​in the parent component.

PR #1633, opened 2024-11-07
fix(ChipList): Minor styling appearance fix

davidkbh:
## **Description**

Changed a negative margin hack into `gap` to prevent bugs where the edge was cut off.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/chips/ChipList.scss
+++ b/projects/novo-elements/src/elements/chips/ChipList.scss
@@ -1,7 +1,7 @@
 @import "../../styles/variables.scss";
 @import "../common/typography/text.mixins";
 
-$novo-chip-list-margin: 4px;
+$novo-chip-list-margin: 8px;
 $novo-chip-input-width: 100px;
 $novo-chip-input-margin: 4px;
 
--- a/projects/novo-elements/src/elements/chips/ChipList.scss
+++ b/projects/novo-elements/src/elements/chips/ChipList.scss
@@ -45,6 +45,6 @@ novo-field.novo-focused {
 
 input.novo-chip-input {
   width: 20px;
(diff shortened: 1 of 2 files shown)


monroepe (dismissed):

projects/novo-elements/src/elements/chips/ChipList.scss
```diff
@@ -7,20 +7,14 @@ $novo-chip-input-margin: 4px;
...
   flex-direction: row;
   flex-wrap: wrap;
   align-items: center;
-  margin: -$novo-chip-list-margin;
-
-  input.novo-input-element,
-  .novo-chip {
-    margin: $novo-chip-list-margin;
-  }
+  gap: $novo-chip-list-margin;
```
dvoegelin, line 17: looks like this change bumped up the input text 4px, misaligning it with the text in other components. you can reproduce this in the query builder demo. we should try to fix this.
4px margin

davidkbh: Thanks for the attention to detail. I will take a look at this and find a clean fix.

This is mainly coming from an effort on my part to get away from a negative margins hack in the chip list that was having other effects. Hopefully we can keep going on it...

dvoegelin (dismissed):

projects/novo-elements/src/elements/chips/ChipList.scss
dvoegelin: fix looks good to me 👍

ndickerson (approved):

monroepe (approved):

PR #1659, opened 2025-03-27
fix(Select): Changes to better support screen readers in <novo-select>

davidkbh:
## **Description**

Sample changes to provide better screenreader accessibility on the <novo-select> component.

#### **Verify that...**

- [x] `npm run lint` passes

- [x] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/common/typography/label/label.component.ts
+++ b/projects/novo-elements/src/elements/common/typography/label/label.component.ts
@@ -1,5 +1,5 @@
 // NG2
-import { Component } from '@angular/core';
+import { Component, HostBinding, input } from '@angular/core';
 import { NovoBaseTextElement } from '../base/base-text.component';
 
 /**
@@ -13,6 +13,8 @@ import { NovoBaseTextElement } from '../base/base-text.component';
  * <novo-label color="grapefruit">Label</novo-label>
  */
 
+let nextId = 0;
+
 @Component({
   selector: 'novo-label,[novo-label]',
   template: ` <ng-content></ng-content> `,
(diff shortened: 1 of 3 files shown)


monroepe (approved):

projects/novo-elements/src/elements/select/Select.scss
```diff
@@ -9,6 +9,18 @@
...
   cursor: pointer;
 
+  &:focus .novo-select-trigger {
+    border-bottom: 1px solid $positive;
+    i {
+      color: rgba(0, 0, 0, 0.73);
+    }
+  }
+  /* cdk-mouse-focused will only apply this state when the user focuses with the mouse.
+  &.cdk-mouse-focused {*/
```
dvoegelin, line 19: if we're not going to use this can we remove it?

projects/novo-elements/src/elements/common/typography/label/label.component.ts
```diff
@@ -21,4 +23,13 @@ import { NovoBaseTextElement } from '../base/base-text.component';
...
   },
 })
-export class NovoLabel extends NovoBaseTextElement {}
+export class NovoLabel extends NovoBaseTextElement {
+  @HostBinding('attr.id')
+  public id: string;
+
+  inputId = input(null, { alias: 'id' });
+
+  ngOnInit() {
```
dvoegelin, line 32: it's generally good practice to implement the OnInit interface on the class for type checking, but we also might be able to eliminate this entirely by just making the `id` property a computed signal and then referencing it on a new `host` property in the component decorator.

monroepe (approved):

PR #1667, opened 2025-05-07
fix(Select): Hide legacy options select if text-filtered

davidkbh:
## **Description**

When we apply text filtration of <novo-select> options, it can cause the UI to mistakenly show the current value as a "legacy option" - a disabled entry that shows it can no longer be selected. Since these are less useful when the set of options is large, we can just hide it. We've better clarified the code that causes these to display, and provided an example for users. There's also an added option allowing legacy options to be hidden completely in any case.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/select/Select.spec.ts
+++ b/projects/novo-elements/src/elements/select/Select.spec.ts
@@ -1,18 +1,51 @@
 // NG
-import { async, TestBed } from '@angular/core/testing';
+import { async, ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
 // App
 import { NovoLabelService } from 'novo-elements/services';
+import { NovoOption, HasOverlay, NovoOverlayModule, NovoOptionModule } from 'novo-elements/elements/common';
 import { Key } from 'novo-elements/utils';
 import { NovoSelectElement } from './Select';
 import { NovoSelectModule } from './Select.module';
+import { SelectionModel } from '@angular/cdk/collections';
+import { ActiveDescendantKeyManager } from '@angular/cdk/a11y';
+import { TAB } from '@angular/cdk/keycodes';
+import { Component, viewChild } from '@angular/core';
 
-xdescribe('Elements: NovoSelectElement', () => {
-  let fixture;
-  let comp;
+@Component({
+  selector: 'test-select-component',
+  template: `
+  <novo-select [value]="value">
+    <novo-option *ngFor="let option of options" [value]="option.value">
+      {{ option.label }}
+    </novo-option>
+  </novo-select>`
+})
+class TestSelectComponent {
+  select = viewChild(NovoSelectElement);
+  options = [{
+    label: 'Option 1',
+    value: '111'
+  }, {
+    label: 'Option 2',
+    value: '222'
+  }, {
+    label: 'Option 3',
+    value: '333'
+  }];
+  value: any;
+}
+
+describe('Elements: NovoSelectElement', () => {
+  let fixture: ComponentFixture<NovoSelectElement>;
+  let comp: NovoSelectElement;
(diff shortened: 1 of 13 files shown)


projects/novo-elements/src/elements/select/Select.ts
```diff
@@ -262,7 +272,7 @@ export class NovoSelectElement
   onModelTouched: Function = () => {};
   filterTerm: string = '';
   filterTermTimeout;
-  filteredOptions: any;
```
dvoegelin, line 265: updating to stricter typing on a public property could cause backwards compatibility issues. can we set this back to just `any`?

projects/novo-elements/src/elements/select/Select.spec.ts
```diff
@@ -35,306 +75,244 @@ xdescribe('Elements: NovoSelectElement', () => {
...
       mockEvent.key = Key.Tab;
+      mockEvent.keyCode = TAB;
+      comp.openPanel();
       comp._handleKeydown(mockEvent);
       expect(comp.overlay.closePanel).toHaveBeenCalledTimes(2);
     });
-    it('should save header', () => {
-      const mockEvent: any = { key: Key.Enter };
+    // Possibly broken?
+    xit('should save header', () => {
```
monroepe, line 199: Did you track down if this is broken or not? It doesnt look like it was disabled before and we shouldn't merge in any disabled tests.

projects/novo-elements/src/elements/select/Select.spec.ts
```diff
@@ -35,306 +75,244 @@ xdescribe('Elements: NovoSelectElement', () => {
...
+      tick();
+      expect(select.contentOptions.length).toBe(2);
+      expect(select.viewOptions.length).toBe(1);
+      const legacyOption: NovoOption = select.viewOptions.get(0);
+      expect(legacyOption.disabled).toBeTruthy();
+      expect(legacyOption.viewValue).toBe('333');
+    }));
+
+    // Expected, but currently broken (may be a niche situation)
+    xit('should present a disabled "legacy option" when updating the list of options (via input) to remove a previously valid value', () => {
```
monroepe, line 274: I am against disabling tests even with a comment. If this is expected behavior then should we just delete the test?

projects/novo-elements/src/elements/select/Select.ts
davidkbh: That does seem like a distant possibility, but yeah, even so it might be better to save for a more major release.

I'll point out that this value seems to be intended only for local and template use - and would generate many errors if it's ever anything other than an array. Still, nothing immediately gained from changing the type now, so I can do that.

projects/novo-elements/src/elements/select/Select.spec.ts
davidkbh: As it happens, this is expected behavior - and I'm actually only leaving this test disabled because the functionality is broken (but, as the current bug was these legacy options showing too often, that may be a very niche error case. Didn't want to trigger too many changes at once)

projects/novo-elements/src/elements/select/Select.spec.ts
```diff
@@ -35,306 +75,244 @@ xdescribe('Elements: NovoSelectElement', () => {
...
+      fixture.componentRef.setInput('hideLegacyOptions', true);
+      fixture.detectChanges();
+      comp.writeValue('bif');
+      fixture.detectChanges();
+      comp.openPanel();
+      expect(comp.viewOptions.length).toBe(3);
+    });
+  })
+
   xdescribe('Function: scrollToSelected', () => {});
```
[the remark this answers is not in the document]
davidkbh: Will do.

monroepe (approved):

PR #1703, opened 2025-09-15
feat(GooglePlacesService): add postal_codes to AddressData for better locality support

vladyslavsemenets:
## **Description**

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


projects/novo-elements/src/elements/query-builder/condition-definitions/address-condition.definition.ts
```diff
@@ -175,6 +175,7 @@ export class NovoDefaultAddressConditionDef extends AbstractConditionFieldDef im
       formatted_address: event.formatted_address,
       geometry: event.geometry,
       name: event.name,
+      postal_code: event?.postal_code,
```
davidkbh, line 178: If event was null here, then `event.name/event.geometry` would already be erroring out.

davidkbh (dismissed):

davidkbh (approved):

PR #1706, opened 2025-09-24
build!: upgrade to Angular 19

ndickerson:
## **Description**

BREAKING CHANGE: Upgrade Angular 17 to 19

This update includes: 
 - Updated Angular core, CLI, and related packages to v19
 - Migrated deprecated APIs and breaking changes
 - Updated TypeScript to minimum required version
 - Many updates to third-party dependencies

See Angular 19 migration guide for full details on breaking changes.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


projects/novo-elements/src/elements/radio/Radio.ts
```diff
@@ -98,7 +99,7 @@ export class NovoRadioElement implements ControlValueAccessor, OnInit {
   }
 
   @Input()
-  get value(): boolean {
```
davidkbh, line 101: Any tests as to what happens if someone assigns this incorrectly?
```
<novo-radio value="false">
```
That would likely evaluate to the string "false", which is then treated as true in boolean logic (eg, when applied to the 'checked' variable)
Usually, the simplest solution is making use of the @BooleanInput decorator, which will handle the logic for you. It also lets you use HTML5 syntax by giving a blank value to mean true. (In the style of; `<input type="checkbox" checked>`)

monroepe (approved):

PR #1773, opened 2026-02-26
chore(Upgrade)!: Angular 20 upgrade

ndickerson:
## **Description**

Upgrade to Angular 20 and Node 22.

This brings Novo Elements up to version 12 because of the breaking change of the Angular and Node versions in use.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


davidkbh (dismissed):

projects/novo-elements/src/elements/agenda/month/AgendaMonthView.ts
```diff
@@ -11,7 +11,7 @@ import {
   Output,
   TemplateRef,
 } from '@angular/core';
-import { addSeconds, differenceInSeconds, getDate, getMonth, getYear, setDate, setMonth, setYear } from 'date-fns';
+import { addSeconds, Day, differenceInSeconds, getDate, getMonth, getYear, setDate, setMonth, setYear } from 'date-fns';
```
monroepe, line 14: any idea how this was working before without the input? i see its used so thats kind odd.

ndickerson: So, this Day is a type that is exported by the date-fns types file. In Angular 19, this is resolved without having to explicitly state where the type comes from, but Angular 20 is more strict and requires that we explicitly import types instead of letting them resolve to whatever type happens to be found.

projects/novo-elements/src/elements/select/Select.spec.ts
```diff
@@ -251,14 +252,17 @@ describe('Elements: NovoSelectElement', () => {
       expect(legacyOption.viewValue).toBe('bif');
     });
 
-    it('should present a disabled "legacy option" when updating the list of options (via content children) to remove a previously valid value', fakeAsync(() => {
+    xit('should present a disabled "legacy option" when updating the list of options (via content children) to remove a previously valid value', fakeAsync(() => {
```
monroepe: i assume these are disabled for a reason (unless you were gonna circle back on them). could you maybe add a quick comment on why they are disabled like what you think the issue is?

monroepe (dismissed):

projects/novo-elements/src/elements/select/Select.spec.ts
ndickerson: Good call, just re-enabled the disabled tests, and was able to also fix an existing disabled one as well

monroepe (dismissed):

monroepe (dismissed):

charlesabarnes (dismissed):

projects/novo-elements/src/elements/simple-table/table.ts
```diff
@@ -80,7 +94,7 @@ export class NovoActivityTableNoResultsMessage {}
     <header *ngIf="(!(dataSource?.totallyEmpty && !state.userFiltered) && !loading) || forceShowHeader">
       <ng-content select="[novo-activity-table-custom-header]"></ng-content>
       <novo-search
-        alwaysOpen="true"
+        [alwaysOpen]=true
```
davidkbh, line 97: Not a showstopper but it might be nice to make this a boolean attribute (so users can just put `alwaysOpen` in the HTML)

davidkbh (dismissed):

charlesabarnes (approved):

davidkbh (approved):

PR #1782, opened 2026-03-27
fix(Memory): Change modalService and toastService for memory optimization

davidkbh:
## **Description**

Do not hold ViewContainerRefs longer than needed. Added a function on toastService: ownViewContainer

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/toast/ToastService.ts
+++ b/projects/novo-elements/src/elements/toast/ToastService.ts
@@ -1,5 +1,5 @@
 // NG2
-import { Injectable } from '@angular/core';
+import { DestroyRef, Injectable, ViewContainerRef } from '@angular/core';
 import { ComponentUtils } from 'novo-elements/services';
 // APP
 import { NovoToastElement } from './Toast';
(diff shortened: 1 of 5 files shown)


projects/novo-elements/src/elements/toast/ToastService.ts
```diff
@@ -31,9 +31,20 @@ export class NovoToastService {
   constructor(private componentUtils: ComponentUtils) {}
 
   set parentViewContainer(view) {
+    console.warn('parentViewContainer is deprecated. Use ownViewContainer(view, destroyRef) to promote lifecycle management');
```
ndickerson, line 34: Just checking - are we going to migrate all locations in Novo as well where this is currently called?

davidkbh: Yup, that's the second part of my changes. It shouldn't cause direct problems if anything still uses this setter, but we should have it cleared away.

monroepe (approved):

ndickerson (approved):

PR #1828, opened 2026-05-27
feat(Chips): add classFunction config for data-driven chip styling

JorgeBerroa:
## **Description**

Adds a new classFunction config option to novo-chips. When provided on a picker's source config, the wrapper runs it with the chip's underlying value and applies the returned class(es) to the rendered <novo-chip> element.

#### **Verify that...**

- [x] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**
image


projects/novo-elements/src/elements/chips/Chip.ts
```diff
@@ -175,6 +176,10 @@ export class NovoChipElement extends NovoChipMixinBase implements FocusableOptio
   @ContentChild(NovoChipRemove) removeIcon: NovoChipRemove;
 
   @Input() type: string;
+  /** Source config object that may contain classFunction */
+  @Input() source: any;
+  /** Label displayed in the chip (used by classFunction) */
+  @Input() label: string;
```
davidkbh: Currently, chips have their label defined through their content. eg,
`<novo-chip><span class="chip-label">{{ item.label }}</span></novo-chip>`

If you're relying on label to be set separately, that's going to lead to a lot of diverging behavior. Is there a specific reason why the classFunction needs that to be passed in? Why would the chip's class be derived from its label (which might be en francais, or escribido en espanol)?

projects/novo-elements/src/elements/chips/Chip.ts
```diff
@@ -259,6 +264,35 @@ export class NovoChipElement extends NovoChipMixinBase implements FocusableOptio
     return this.selectable && (this._chipListMultiple || this.selected) ? this.selected.toString() : null;
   }
 
+  /**
+   * Dynamic classes from `source.classFunction`. Normalizes the function's return
+   * value (string, string[], or {[className: string]: boolean}) into a
+   * space-separated class string suitable for the host element's `[class]` binding.
+   */
+  get dynamicClasses(): string {
```
davidkbh: Why do you need this function to split objects into string pieces? Why not use ngClass?

davidkbh (approved):

leemarscott (approved):

PR #1846, opened 2026-06-24
fix(Address): harden Google Maps lazy-load and flat address shape (BH…

ndickerson:
…-88540)

Address code-review findings on the Maps SDK lazy-load / address-search-service work:

- Render flat (address-search-service) chips: add addressLabel() so the chip falls back to formattedAddress when formatted_address is absent, fixing blank chips on the server path.
- Resolve loadGoogleMaps via google.maps.importLibrary('places') on script load so the promise settles only once the places library is usable (loading=async fires onload before it is ready).
- Convert getListQuery/getCurrentLocationInfo/getPlaceLocationInfo to async/await with try/catch: clear the current-location spinner in finally, surface an empty list on prediction-load failure, and add the missing SDK guard before getGeoPlaceDetail.
- Drop undefined Maps loader params instead of serializing them as "undefined", and warn when a second, different googleApiKey is requested (Maps JS can only load once per page).
- Demo: route googlePlacesKey through NOVO_ADDRESS_CONFIG so the address components exercise the real lazy-load path instead of a duplicated bootstrap script.

Adds tests covering importLibrary sequencing, undefined-param filtering, key-conflict warning, async error/finally paths, and addressLabel.

## **Description**

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/query-builder/condition-definitions/address-condition.definition.ts
+++ b/projects/novo-elements/src/elements/query-builder/condition-definitions/address-condition.definition.ts
@@ -192,7 +192,23 @@ export class NovoDefaultAddressConditionDef extends AbstractConditionFieldDef im
           place_id: event.place_id,
           types: event.types,
         }
-      : { ...event };
+      : {
+          address1: event.address1,
+          address2: event.address2,
+          city: event.city,
+          state: event.state,
+          zip: event.zip,
+          countryName: event.countryName,
+          countryCode: event.countryCode,
+          formattedAddress: event.formattedAddress,
+          location: event.location,
+          viewport: event.viewport,
+          referenceId: event.referenceId,
+          postalCodes: event.postalCodes,
+          postal_codes: event.postal_codes,
+          types: event.types,
+          radius: event.radius,
+        };
     const current: AddressData | AddressData[] = this.getValue(formGroup);
     const updated: AddressData[] = Array.isArray(current) ? [...current, valueToAdd] : [valueToAdd];
     formGroup.get('value').setValue(this.updateRadiusInValues(formGroup, updated));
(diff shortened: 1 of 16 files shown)


davidkbh (dismissed):

projects/novo-elements/src/elements/query-builder/condition-definitions/address-condition.definition.ts
```diff
@@ -121,6 +122,9 @@ export class NovoDefaultAddressConditionDef extends AbstractConditionFieldDef im
   private _addressChangesSubscription: Subscription = Subscription.EMPTY;
 
   public element = inject(ElementRef);
+  // App-wide address config; when provided with the server path, the picker uses the address-search-service
+  // instead of the Google client SDK. Null leaves the picker on its Google default.
+  protected addressConfig: PlacesSettings = inject(NOVO_ADDRESS_CONFIG, { optional: true });
```
monroepe: could this be marked as readOnly?

projects/novo-elements/src/elements/query-builder/condition-definitions/address-condition.definition.ts
```diff
@@ -64,7 +64,7 @@ import { NovoSelectElement } from 'novo-elements/elements/select';
           <novo-field #novoField class="address-location">
             <novo-chip-list [(ngModel)]="chipListModel" [ngModelOptions]="{ standalone: true }" (click)="openPlacesList(viewIndex)">
               <novo-chip *ngFor="let item of formGroup.get('value').value" (removed)="remove(item, formGroup, viewIndex)">
-                <novo-text ellipsis [tooltip]="item.formatted_address" tooltipOnOverflow>{{ item.formatted_address }}</novo-text>
+                <novo-text ellipsis [tooltip]="addressLabel(item)" tooltipOnOverflow>{{ addressLabel(item) }}</novo-text>
```
monroepe: could we create a pipe for addressLabel instead of the function call?

monroepe (approved):

PR #1852, opened 2026-07-03
fix(DataTableCheckboxHeaderCell): Bug with selecting all new page rows when paginating

tfechtnerbh:
## **Description**

Fix a bug with selecting all new page rows when paginating to a new page and 'all match records' is selected

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


monroepe (approved):

davidkbh (approved):
I'm just a little concerned about a potential infinite loop from selectRows calling selectRows, but I'll defer to Peter's expertise given that it sounds like this affects an active issue.

PR #1869, opened 2026-07-22
feat(bh2026): tokenize component styles and update form styles

ndickerson:
Replace raw CSS values with bh2026 design tokens across 23 component stylesheets. All form-adjacent display text (labels, hints, field-message, character/record counts, description, column headers) now follows a consistent size/weight/color hierarchy. Picker results corner-clipping fixed by moving the border onto the host element.

Tokenization changes:
- bh2026.scss: strip raw fallbacks from background, text, border, button, and font-size semantic vars; add spacing-scale overrides for the 16px rem base
- tabs/tab.scss: primary color → var(--button-primary-background); font-weight → tab tokens
- tiles/Tiles.scss: border and muted-text raw values → tokens
- card/Card.scss: spacing, shadow, and header typography → tokens
- header/Header.scss, toolbar: box-shadow rgba → var(--color-transparency-black-10)

Form text hierarchy (bh2026):
- field/field.scss: label → input-label-sm tokens + weight-600; value → input-value tokens
- field/hint/hint.scss: override --font-size-caption (1.4rem) down to --typography-font-size-12
- Form.scss: floating label active states, static control label, input value, field-message, character-count, record-count, error-message character-count — all explicitly overridden ($grey is a compile-time SCSS variable; CSS custom property cascade cannot reach it)
- ControlGroup.scss: section description label and column-header spans → decoration scale

Picker results:
- PickerResults.scss: bh2026 block moves border to picker-results host element so overflow:hidden + border-radius does not clip child element border corners

## **Description**

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


projects/novo-elements/src/elements/field/field.scss
```diff
@@ -109,15 +109,19 @@
...
+      font-weight: var(--typography-font-weight-400);
+      font-size: var(--typography-input-label-sm-font-size);
+      letter-spacing: var(--typography-input-label-sm-letter-spacing);
+      color: var(--color-dark);
+    }
   }
   .novo-field-input ::ng-deep .novo-input-element {
-    font-size: 1.4rem;
-    font-weight: 500;
+    font-size: var(--typography-input-value-default-font-size);
```
davidkbh, line 124: Is this a variable name we can shorten, especially if we're using it in a lot of places?

davidkbh (dismissed):

projects/novo-elements/src/elements/field/field.scss
ndickerson: We can, that would be a shortening within the NE bh2026 stylesheet. I am not going to shorten that right now, since the styles are too much in flux from the design tokens up. But once we have all finalized design tokens, we can take a look at what shorter names we can use as aliases from within Novo Elements. I'm afraid to add too many additional styles at the NE level at this point since I'm not sure what will change next.

monroepe (dismissed):

monroepe (dismissed):

monroepe (approved):

charlesabarnes (approved):

PR #1879, opened 2026-07-30
fix(OutsideClick): Fix event triggers on base class

davidkbh:
## **Description**

This base class was only usable within TypeScript calls, and couldn't be used in Angular templates.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**


davidkbh:
This has been wrapped into another MR.

PR #1896, opened 2026-08-11
feat(): Create new icon button option for tip-well

leah-arden:
## **Description**

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

image


projects/novo-elements/src/elements/tip-well/TipWell.scss
```diff
@@ -19,7 +19,7 @@
         flex-shrink: 0;
         text-align: center;
         margin-right: var(--spacing-md);
-        color: #aaaaaa;
+        color: var(--color-text-subtle, #aaaaaa);
```
davidkbh: Is this just for the new theme, or for all tipwells?

davidkbh (dismissed):

projects/novo-examples/src/components/tip-well/icon-button-tip-well/icon-button-tip-well-example.html
```diff
@@ -0,0 +1,13 @@
+<novo-tip-well
+  data-automation-id="icon-button-tip-well"
+  name="IconButtonDemo"
+  icon="info"
+  buttonTheme="icon"
+  buttonSize="small"
+  buttonText="&#215;"
```
ndickerson: Why is the button text being set to this? In the tip well it's normally an "Ok, Got it".

Is this replacing that with an icon instead?

leah-arden: The ask from UX is to have the tipwell have an "x" instead of an "okay, got it". I can have it use an icon for the example if that is clearer.

something like this:

image

projects/novo-elements/src/elements/tip-well/TipWell.scss
leah-arden: I updated this to just be for the new theme

davidkbh:
I suggested using an input name other than "theme" to avoid confusion but other than that, looks good.

davidkbh (dismissed):

ndickerson (approved):

davidkbh (approved):

PR #1915, opened 2026-09-04
fix(Form): revert @for to *ngFor in DynamicForm templates

monroepe:
## **Description**

@for with track-by-key reuses component instances across re-renders,
preventing novo-control from reinitializing when the form group is
replaced. *ngFor without trackBy tracks by identity, ensuring components
are recreated and ngOnInit runs fresh on each form rebuild.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/STANDARDS.md
+++ b/STANDARDS.md
@@ -150,8 +150,9 @@ All user-visible strings must go through `NovoLabelService` — never hardcode d
 ### Template Control Flow
 
 -   Use built-in control flow: `@if`, `@for`, `@switch`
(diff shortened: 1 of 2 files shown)


leemarscott (approved):

davidkbh:
Interesting find - maybe this would change if we move the functions of ngOnInit to an initialize function, and run it anytime anything major changes from the inputs. That said, this seems like a good solution for now.

davidkbh (approved):

PR #1916, opened 2026-09-04
feat(Dropdown): Address a niche case with native buttons

davidkbh:
<novo-dropdown> supports a content button to serve as the dropper. It presumes this will be a NovoButtonElement, but doesn't necessarily need to be so. This allows for that possibility.

#### **Verify that...**

- [ ] `npm run lint` passes

- [ ] Run `Novo Automation`

##### **Screenshots**

### DIFF
--- a/projects/novo-elements/src/elements/dropdown/Dropdown.ts
+++ b/projects/novo-elements/src/elements/dropdown/Dropdown.ts
@@ -7,17 +7,20 @@ import {
   AfterViewInit,
   ChangeDetectorRef,
   Component,
-  ContentChild,
+  computed,
+  contentChild,
   ContentChildren,
   Directive,
   ElementRef,
   EventEmitter,
   HostListener,
+  inject,
   Input,
   OnDestroy,
   OnInit,
   Output,
   QueryList,
+  Renderer2,
   ViewChild,
 } from '@angular/core';
 // Vendor
@@ -26,6 +29,8 @@ import { takeUntil } from 'rxjs/operators';
 // App
 import { NovoButtonElement } from 'novo-elements/elements/button';
 import {
+  _countGroupLabelsBeforeOption,
+  _getOptionScrollPosition,
   CanDisableCtor,
   HasOverlayCtor,
   HasTabIndexCtor,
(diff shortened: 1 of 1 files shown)


monroepe (dismissed):

monroepe (dismissed):

projects/novo-elements/src/elements/dropdown/Dropdown.ts
```diff
@@ -175,9 +178,10 @@ export class NovoDropdownElement extends NovoDropdownMixins implements OnInit, A
   }
 
   public ngAfterContentInit(): void {
+    const button = this.button();
```
ndickerson, line 181: What about if this.button() is undefined? Does that workflow change with these code changes?

```
   const button = this.button(); // undefined
   button.tabIndex = -1;        // TypeError: Cannot set properties of undefined
```

Would it be worth adding a null check or throwing an error with a meaningful message?

projects/novo-elements/src/elements/dropdown/Dropdown.ts
```diff
@@ -222,6 +222,12 @@ export class NovoDropdownElement extends NovoDropdownMixins implements OnInit, A
 
   public set items(items: QueryList<NovoItemElement>) {}
 
+  // If the dropdown is using a basic <button> with no theme, it will not appear in the ContentChild directive.
+  private _findNativeButton() {
+    return Array.prototype.find.call(this.element.nativeElement.children,
```
ndickerson, line 233: A nested button (e.g. inside a wrapper <span>) wouldn't be found here. Would we want to use querySelector('button') to handle nested cases?

projects/novo-elements/src/elements/dropdown/Dropdown.ts
davidkbh: If it's not there, then the component is going to break with an error and be unusable no matter what we do. I could present a more specific error message, at least.

projects/novo-elements/src/elements/dropdown/Dropdown.ts
davidkbh: That is true, but is also true of the ContentChild selector - you'd need to provide the `descendants: true` option to get it working that way. When we configure elements with templates, it's a common pattern to require them to be directly under the component. So you can't write a table header as: `<novo-data-table> ... <div> ... <div> ... <section> ... <data-table-header>`. I think that pattern helps to optimize, and make it less confusing how two elements could connect in an unseen way (eg, an overlay template taking a button 5 levels deeper as its "close" button)

ndickerson (approved):

monroepe (approved):

